PR #24, opened 2015-04-07
Cosmetic changes 

manucorporat:
- DRY. removed a lot of duplicated code in baked_in.go. Easier to read, maintain and debug

- Faster `StructValidationErrors.Error()` by using `bytes.Buffer`

- StructValidationErrors is allocated after some checks, it may save some cycles...

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -54,7 +54,6 @@ func isURI(top interface{}, current interface{}, field interface{}, param string
 }
 
 func isURL(top interface{}, current interface{}, field interface{}, param string) bool {
-
 	st := reflect.ValueOf(field)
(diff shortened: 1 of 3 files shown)


manucorporat:
If you want to merge let me know, so I can fix the unit tests or restore the old error messages.

joeybloggs:
It looks great! I think I will merge it, I can update the unit test if it's too much trouble.

I do have one request though, if you could make the pull request against the v4-development branch instead of v4 directly; I would also like to update a few issues I found in my documentation at the same time.

Thanks!

joeybloggs:
changes merge!

PR #111, opened 2015-07-12
V5 development

zhing:
when validating length of a web string, we care about its character length but not its byte length. Such as the character length of "你好a" is 3 which we demand, but not its byte length 2*3+1 which makes confusion.


joeybloggs:
makes perfect sense I will merge this asap ( most likely this afternoon )

I will also make the same changes to some other methods like:
isGt
isGte
isLt
isLte
... and any others that should also be modified

PR #114, opened 2015-07-19
V6 development

joeybloggs:


doc.go
```diff
@@ -148,6 +78,11 @@ included within the parameter i.e. excludesall=, you will need to use the UTF-8
 representation 0x2C, which is replaced in the code as a comma, so the above will
 become excludesall=0x2C
 
+NOTE3: pipe is the default separator of or validation tags, if you wish to have a pipe
+included within the parameter i.e. excludesall=| you will need to use the UTF-8 hex
+representation 0x7C, which is replaced in the code as a pipe, so the above will
+become excludesall=0x7C
+
```
tschaub: Was this paragraph supposed to replace the one above ("NOTE2: comma is the default separator of validation tags").  Reading through the current docs, it is hard to understand which of `,` or `|` is the default separator.

> Comma (",") is the default separator of validation tags. If you wish to
> have a comma included within the parameter (i.e. excludesall=,) you will need to
> use the UTF-8 hex representation 0x2C, which is replaced in the code as a comma,
> so the above will become excludesall=0x2C.
> 	type Test struct {
> 		Field `validate:"excludesall=,"`    // BAD! Do not include a comma.
> 		Field `validate:"excludesall=0x2C"` // GOOD! Use the UTF-8 hex representation.
> 	}
>
> Pipe ("|") is the default separator of validation tags. If you wish to
> have a pipe included within the parameter i.e. excludesall=| you will need to
> use the UTF-8 hex representation 0x7C, which is replaced in the code as a pipe,
> so the above will become excludesall=0x7C
> 	type Test struct {
> 		Field `validate:"excludesall=|"`    // BAD! Do not include a a pipe!
> 		Field `validate:"excludesall=0x7C"` // GOOD! Use the UTF-8 hex representation.
>   }

joeybloggs: thanks @tschaub I'll correct the docs; next time if you could create a new issue that'd be great as it's confusing to have a discussion in a closed or merge issue/PR :)

PR #154, opened 2015-08-19
Update examples_test.go

ardan-bkennedy:
Please don't use relative paths. I want to use the package but this is causing problems. I plan on using the package during my GolangUK talk on Friday but I need this fixed.


ardan-bkennedy:
Wrong version number

PR #155, opened 2015-08-19
Update examples_test.go

ardan-bkennedy:
This is better. We can't use relative paths for imports. If you can get this change in, it will help with my talk at GolangUK on Friday.

### DIFF
--- a/examples_test.go
+++ b/examples_test.go
@@ -3,7 +3,7 @@ package validator_test
 import (
 	"fmt"
 
-	"../validator"
+	"gopkg.in/bluesuncorp/validator.v6"
 )
 
 func ExampleValidate_new() {


joeybloggs:
No problems, @ardan-bkennedy I only had it that way while adding/updating some new functionality that didn't exist before and so had to point to my fork and not the stable branch; guess I forgot to change it back.

just out of pure curiosity, GolangUK that's awesome how will you be using this library?

P.S. I will ensure that v7 get's released before Friday and some changes from it backported into v6

ardan-bkennedy:
This is great !!.  I am using the package in this code (https://github.com/ArdanStudios/gotraining/tree/master/13-http/api) which is part of my training material. I am giving a talk on dependency management and I am using the code to show for solutions. In vendoring your code and running tests, I came across the problem.

joeybloggs:
That's Awesome! I guess this library is reaching further than I thought

Best of luck at the conference :+1:

PR #166, opened 2015-09-02
Added checking for nil receiver on Validator methods

flowonyx:
If `validator.New` is not called, the various validation methods can still be called since they are on the pointer, but it will panic with a slightly cryptic message. This still panics, but with a clear message that will hopefully save some other developers time in debugging.


joeybloggs:
I should be able to get this done later tonight

flowonyx:
Thanks! I have no problem with you implementing it in any other way, I just thought it was an easy way to add some clarity.

PR #192, opened 2015-10-02
Fix typo on README so the example compiles

fabioxgn:


joeybloggs:
Thanks! Must have been there a while

PR #243, opened 2016-06-21
Fix typo: ne is validation key for not equal

kenkouot:


joeybloggs:
Thanks! Is also a typo in the paragraph above if you would like to also fix and as soon as I'm back at a computer will merge and cut a release.

kenkouot:
Done!

joeybloggs:
Thanks! @kenkouot

PR #256, opened 2016-09-22
Fix universal translator imports so that goimports doesn't get remove them

jawher:
@go-playground/admins

Running goimports on the codebase deletes all the imports of "github.com/go-playground/universal-translator".

The fix consists of explicitly qualifying the import statements with ut.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -6,7 +6,7 @@ import (
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


joeybloggs:
Thanks @jawher I didn't realize that would happen...

I'll merge this as soon as I get a chance this afternoon

Issue #282, opened 2017-06-06
How to get field name


s4l1h:
### Package version eg. v8, v9: 

v9

### Issue, Question or Enhancement:
How to get field name eg: **Name**
### Code sample, to showcase or reproduce:

```go

// Group Models
type Group struct {
	Name        string `json:"name" validate:"unique,required"`
	Description string `json:"Description" validate:"required"`
}

var valid = validator.New()
valid.RegisterValidation("unique", ValidUnique)

func ValidUnique(fl validator.FieldLevel) bool {
        //How to get field name eg: Name
	fmt.Printf("Validating %s\n", fl.Field())



	return false
}
```

joeybloggs:
Hey @s4l1h 

So I'm assuming you are looking to access the struct field name from within the validation, currently there is no method to get that value...bit after a validation fails your can https://github.com/go-playground/validator/blob/v9/examples/simple/main.go#L72

Any reason you need the field name from within the validation?

s4l1h:
check if record exists in database

important problem

```go

// Group Models
type Group struct {
	Name        string `json:"name" validate:"unique,required"`
	Description string `json:"Description" validate:"required"`
}

var valid = validator.New()
valid.RegisterValidation("unique", ValidUnique)

func ValidUnique(fl validator.FieldLevel) bool {

	currentField, _, _ := fl.GetStructFieldOK()
	table := currentField.Type().Name() // table name
	value := fl.Field().String() // value
        column := "????" // need this..
	sql := fmt.Sprintf("select count(*) from %s where %s='%s'", table, column, value)
	if Count(sql) { // check database...
		return false
	}
	return true
}
```

joeybloggs:
@s4l1h makes perfect sense

I think I can pass the cached filed info along and then add a couple of methods to the FieldLevel interface to access both the struct and alt name's.

It may take me a few days, I'm out of town right now, but will try to make the changes asap.

s4l1h:
@joeybloggs thanks for reply and effort.I can help.
Pls check pr.

PR #283, opened 2017-06-07
added two new methods. CTag and CField for FieldLevel #282

s4l1h:
Fixes Or Enhances # .

Change Details:

- 
- 
- 

@go-playground/admins

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -82,6 +82,16 @@ type cField struct {
 	cTags      *cTag
 }
 
+// Name return field name
+func (cf *cField) Name() string {
+	return cf.name
+}
+
+// AltName return field altname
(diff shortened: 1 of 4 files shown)


joeybloggs:
Thanks @s4l1h  but I'd like to expose that information a little differently without exposing cField nor cTag

I appreciate the help, but may close the PR after implementing

joeybloggs:
Update see: https://github.com/go-playground/validator/tree/add-fl-field-name

just have to write the tests...they need to cover not only a normal FieldLevel, but also and array and map to ensure the names are appropriate with key and or index included.

s4l1h:
@joeybloggs pretty good .Maybe you can add another function direct return **Struct Name**

PR #305, opened 2017-08-30
Don't include type name in ReportValidationErrors

aaronlehmann:
Using ReportValidationErrors produces a distorted field path because the
type name from the errors is included. This can cause a "namespace"
like:

    Struct1Name.Field1Name.Field2Name.Struct2Name.Field3Name

This change strips out the type name from the errors when they are being
appended to an existing namespace.

I would prefer for the struct name to never be included (I have to work around this in my application), but I suspect this would be a more disruptive change.

@go-playground/admins

### DIFF
--- a/struct_level.go
+++ b/struct_level.go
@@ -3,6 +3,7 @@ package validator
 import (
 	"context"
 	"reflect"
+	"strings"
 )
 
 // StructLevelFunc accepts all values needed for struct level validation
@@ -170,8 +171,33 @@ func (v *validate) ReportValidationErrors(relativeNamespace, relativeStructNames
 	for i := 0; i < len(errs); i++ {
 
 		err = errs[i].(*fieldError)
-		err.ns = string(append(append(v.ns, relativeNamespace...), err.ns...))
-		err.structNs = string(append(append(v.actualNs, relativeStructNamespace...), err.structNs...))
+
+		ns := err.ns
+		prefix := append(v.ns, relativeNamespace...)
+		if len(prefix) > 0 {
+			// Strip out type name
+			idx := strings.Index(ns, ".")
+			if idx == -1 {
+				ns = ""
+			} else {
+				ns = ns[idx+1:]
+			}
+		}
+		err.ns = string(append(prefix, ns...))
+
+		structNs := err.structNs
+		structPrefix := append(v.actualNs, relativeStructNamespace...)
+		if len(structPrefix) > 0 {
+			// Strip out type name
+			idx := strings.Index(structNs, ".")
+			if idx == -1 {
+				structNs = ""
+			} else {
+				structNs = structNs[idx+1:]
+			}
+		}
+
+		err.structNs = string(append(structPrefix, structNs...))
 
 		v.errs = append(v.errs, err)
 	}
--- a/validator_test.go
+++ b/validator_test.go
@@ -279,24 +279,24 @@ func StructValidationTestStructReturnValidationErrors(sl StructLevel) {
 
 	s := sl.Current().Interface().(TestStructReturnValidationErrors)
 
-	errs := sl.Validator().Struct(s.Inner1.Inner2)
+	errs := sl.Validator().Struct(s.Inner1.Inner2Field)
 	if errs == nil {
 		return
 	}
 
-	sl.ReportValidationErrors("Inner1.", "Inner1.", errs.(ValidationErrors))
+	sl.ReportValidationErrors("Inner1.Inner2Field.", "Inner1.Inner2Field.", errs.(ValidationErrors))
 }
 
 func StructValidationTestStructReturnValidationErrors2(sl StructLevel) {
 
 	s := sl.Current().Interface().(TestStructReturnValidationErrors)
 
-	errs := sl.Validator().Struct(s.Inner1.Inner2)
+	errs := sl.Validator().Struct(s.Inner1.Inner2Field)
 	if errs == nil {
 		return
 	}
 
-	sl.ReportValidationErrors("Inner1JSON.", "Inner1.", errs.(ValidationErrors))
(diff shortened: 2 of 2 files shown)


joeybloggs:
Hey @aaronlehmann would you be able to produce an example outlining the issue, I'm not quite sure I fully have my head wrapped around the issue and why adjusting the relative namespace wouldn't work?

aaronlehmann:
I encountered the issue when I needed to wrap part of the validation of a struct to make the validation conditional. The data structures look sort of like this:

```go
type Model struct {
        Items []*Item `json:"items"`
        ...
}

type Item struct {
        Validate bool
        Str1 `validate:required`
        ...
}
```

Basically, I only want to validate the items in `Items` that have `Validate` set to true. Because of this, I can't use `dive`. So what I'm doing is registering a struct validation with this logic:

```go
func init() {
        validator.RegisterStructValidation(validateModel, Model{})
}

func validateModel(sl validator.StructLevel) {
        model := sl.Current().Interface().(Model)
        for i, item := range quote.Items {
                if !item.Validate {
                        continue
                }

                err := sl.Validator().Struct(item)
                if errs, ok := err.(validator.ValidationErrors); ok {
                        sl.ReportValidationErrors(fmt.Sprintf("items[%d].", i), fmt.Sprintf("Items[%d].", i), errs)
                }
        }
```

In a request handler, I call `validator.Struct()` on a `Model`. I get back errors with namespace values like `Model.Items[0].Item.Str1`, but would expect `Model.Items[0].Str1`. The additional type name that gets inserted doesn't make sense and can't really be filtered out on my side.

The changes here check if there's a namespace prefix in `ReportValidationErrors`, and if there is, it strips out the struct name from the errors, so we don't end up with a namespace that contains an unexpected struct name in the middle.

For me it would be ideal to never include any struct names in the "namespace" value. I only want to report the fields with problems to my users, not my type names (which are internal). But it's easy for me to strip out the first item in the namespace string as a type, and with this change I can keep on doing that.

joeybloggs:
@aaronlehmann sorry for my late replies, been getting all settled after my move.

I see how this would be useful, never expected anyone to use it that way, but makes perfect sense :)

I do have some concerns about a couple pieces of code will comment on in the review; and if you could provide test for 100% test coverage that would also be great!

struct_level.go
```diff
@@ -170,8 +171,33 @@ func (v *validate) ReportValidationErrors(relativeNamespace, relativeStructNames
...
 
 		err = errs[i].(*fieldError)
-		err.ns = string(append(append(v.ns, relativeNamespace...), err.ns...))
-		err.structNs = string(append(append(v.actualNs, relativeStructNamespace...), err.structNs...))
+
+		ns := err.ns
+		prefix := append(v.ns, relativeNamespace...)
+		if len(prefix) > 0 {
+			// Strip out type name
+			idx := strings.Index(ns, ".")
```
joeybloggs: Is this going to work when it's multidimensional? eg your example is `Model.Items[0].Items.Str`

What if it's two levels deep eg. `Model.Model2.Items[0].Items.Str`, this code looks like it will cut off `Model2`? perhaps this should be `LastIndex`?

aaronlehmann: I don't think that multiple dimensions would cause any issues. Every time `Struct` gets called, it prepends the struct name. The code I've added undoes this prepending when those errors are bubbled up a level with `ReportValidationErrors`. If I understand your example correctly, that would involve multiple nested calls to `Struct`, which use `ReportValidationErrors` inside the validator function. I don't expect any problems from this, but I can write tests when I get a chance.

One thing I now realize this might break, though, is a nested call to `Var`. That wouldn't prepend a type name, so stripping out the first item from the namespace would be wrong.

Perhaps `Var` would be a way to work around my original problem, without any of these changes? I suppose I could call `Var` instead of struct, and that would produce errors without a type name attached.

mvrhov:
This is what also bothers me. The type information is redundant. Probably in almost all cases.
E.g If you validate the struct that you fill from form. Then the easiest mapping of Errors to form fields is just the Field name.

joeybloggs:
@mvrhov if I understand you correctly, the problem then becomes the consistency of how errors are used/handled, the top level ones don't have type info, but nested ones do?

IMO this would only be a sources of confusion and pain for those wanting to write wrappers around handling of the errors.

mvrhov:
There is no type info anywhere in the chain. Just the fieldds

joeybloggs:
@mvrhov I must not be understanding correctly, can you provide an example?

mvrhov:
If we take the structs from example simple, then I'd expect the following:

I'll use json notation:
```json
{
"FirstName": "error string",
"Addresses[X].Street":  "error string",
"Addresses[Y].Planet":  "error string",
}
```
And not:
```json
{
"User.FirstName": "error string",
"User.Addresses[X].Address.Street":  "error string",
"User.Addresses[Y].Address.Planet":  "error string",
}
```

```go
// User contains user information
type User struct {
	FirstName      string     `validate:"required"`
	LastName       string     `validate:"required"`
	Age            uint8      `validate:"gte=0,lte=130"`
	Email          string     `validate:"required,email"`
	FavouriteColor string     `validate:"iscolor"`                // alias for 'hexcolor|rgb|rgba|hsl|hsla'
	Addresses      []*Address `validate:"required,dive,required"` // a person can have a home and cottage...
}

// Address houses a users address information
type Address struct {
	Street string `validate:"required"`
	City   string `validate:"required"`
	Planet string `validate:"required"`
	Phone  string `validate:"required"`
}
```

deankarn:
@mvrhov I see what you mean, I have it this way due to type `interface{}` where the type is not guaranteed then the error will allow differentiation of the two, especially if they have the same field names.

Perhaps it's overkill though, I'll add a ticket about possibly removing it for v10, until then going to close this out as it'd be a breaking change right now.

PR #340, opened 2018-02-02
Support translating field errors when field is of kind reflect.Ptr

glb:
Enhances translated strings for fields for kind `reflect.Ptr`.

Change Details:
- Previously, if fields were of kind `reflect.Ptr`, the translation of validation errors used the default string, resulting in unexpected output. Now the default translations use the field's `Elem()` kind (for example for `*string` the `Elem()` is `string`), resulting in much more natural output.

@go-playground/admins


joeybloggs:
Thanks @glb

glb:
Thank you!

PR #343, opened 2018-02-16
Add oneof validation

ansel1:

Change Details:

Validates that a string or number value equals one of a set of allowed values.  Examples:

    validator.Var(“red”, “oneof=red green”)
    validator.Var(6, “oneof=6 7 8”)

Handles values of type string, int, int8, int16, int32, int64, uint, uint8, uint16, uint32, and uint64.

The translation reads "Color must be one of [red green]".

When the field value is a numeric type, it's converted to a string before comparison.  So the param values are always treated as strings.

The param values are parsed using strings.Fields() (which splits by whitespace).  To speed it up, there is an internal cache of parsed params, protected by a mutex.  Benchmarking demonstrated this cut eval time in half, from 500 ns to 250 ns.  They are both pretty negligible, so if you're prefer less complexity, it's not a big deal to lose the cache.

For comparison, I also benchmarked using the sync.Map type rather than a mutex, but that was slower.

Also benchmarked using slice iteration vs a map lookup for testing the field against the param values.  slice iteration proved faster when the number of param values was 10.  I don't expect more values than that.
- 
- 
- 

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -6,7 +6,9 @@ import (
 	"net"
 	"net/url"
 	"reflect"
+	"strconv"
 	"strings"
+	"sync"
 	"time"
 	"unicode/utf8"
 )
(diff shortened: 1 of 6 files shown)


joeybloggs:
Thanks @ansel1 for the PR! A couple thoughts:
- it seems to be failing the tests in the CI
- have you tried a RWMutex to see what perf may be?
- why not also add float32 and float64 to the supported types :)

ansel1:
I'll try out the RWMutex.  

I didn't do floats because of the precision thing.  Wasn't sure you could reliably match floats to their string representations, unless I truncate them to the string's precision.  Which all seemed overly complicated for a questionable use case.  It didn't seem likely one would want to validate a float value against fixed values (again, because of the precision thing).

Re the CI failure, I think there's an issue in the way you build.  I think the translator/en package is not being tested against the same commit of the validator.v9 package.   The build output shows that the `go get` command in the build sequence is fetching the master branch of `gopkg.in/go-playground/validator.v9` into the GOPATH.  That will be a different version of code than that in `/home/runner/workspace/src/gopkg.in/go-playground/validator.v9`.  I'm guessing `/home/runner/workspace` is not on the GOPATH first.

I can resolve by pushing this PR with the changes to translations/en, then push those in a second PR.  I believe that would build.

joeybloggs:
Thanks @ansel1 

totally agree about float precision. About the build failures thanks for reminding me, I found a way to get around that, just haven't implemented yet.

ansel1:
swapped out the mutex for rwmutex.  does seem to increase concurrent performance, with no perceptible penalty.

joeybloggs:
awesome @ansel1 thanks, will try and get this merged tonight

LjtGentle:
json.Unmarsh to value is map[string]interface number will be float64,so oneof why not support float64?

PR #351, opened 2018-03-28
Add validation for base64 URL safe values

kgeorgiou:

Change Details:

Added a new validation tag, `base64url`, for the Base64 URL safe encoding as specified in [section 5](https://tools.ietf.org/html/rfc4648#section-5) of [RFC4648](https://tools.ietf.org/html/rfc4648).

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -96,6 +96,7 @@ var (
 		"url":              isURL,
 		"uri":              isURI,
 		"base64":           isBase64,
+		"base64url":        isBase64URL,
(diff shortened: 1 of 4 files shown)


joeybloggs:
Thanks @kgeorgiou I'm out of town until next week, will try and merge when I get back

joeybloggs:
looks awesome! @kgeorgiou thanks!

I will adjust the documentation ever so slightly letting people know it's RFC4648 base64 URL encoding

PR #358, opened 2018-05-01
add missing `Engine()` to default validator for gin override example

Ashtonian:

Change Details:

- update the gin override example to include a function `Engine()` to meet the new interface requirements for gin, https://github.com/gin-gonic/gin/blob/master/binding/binding.go#L46 
- unable to write tests because the type of change 

@go-playground/admins


joeybloggs:
Thanks @Ashtonian

Issue #359, opened 2018-05-03
Email - Hostname is not validated correctly


danielpoe:
### Package version eg. v8, v9: 
v9

### Issue, Question or Enhancement:
According to http://www.ietf.org/rfc/rfc1123.txt
Hostnames are limited  to letters-digits-hyphen

Currently underscores are allowed in the validator 

### Code sample, to showcase or reproduce:

```go
Test struct {
Email            string `form:"email" validate:"required,email" conform:"email"`
}

Validation of "test@test_test.org" suceeds but test_test.org is not a valid hostname
```

joeybloggs:
@danielpoe thanks for reporting, I will try and get to this, but might take a little bit, PR's are welcome if you would like it sooner :)

jadhub:
I supplied a PR for this: https://github.com/go-playground/validator/pull/362

PR #362, opened 2018-05-09
Fix for Email - Hostname is not validated correctly #359

jadhub:
Fixes https://github.com/go-playground/validator/issues/359

"Email - Hostname is not validated correctly #359"

While Hostnames could potentially contain underscores, Domains should per RFC not contain underscores.

Tests
- Extended validator_test.go to include a test for a domain with underscores to expectedly return an error. Tests were successful after the change.

Change Details:

- Removed Underscores from two occurences in Alternations in the Regex.

@go-playground/admins


joeybloggs:
awesome @jadhub! thanks will merge asap!

PR #403, opened 2018-10-18
Fixed UUID validation regexs

### LINKED ISSUE #402: Incorrect UUID validation regex

v9

The regular expression validation for UUIDs only allows lowercase characters.

[RFC 4122](http://www.ietf.org/rfc/rfc4122.txt) allows uppercase characters for input of functions but requires lowercase characters on output:

>  Each field is treated as an integer and has its value printed as a
  zero-filled hexadecimal digit string with the most significant
  digit first.  The hexadecimal values "a" through "f" are output as
  lower case characters and are case insensitive on input.

In addition the acceptable hex digits are (according to section 3 of the RFC):

```
hexDigit =
    "0" / "1" / "2" / "3" / "4" / "5" / "6" / "7" / "8" / "9" /
    "a" / "b" / "c" / "d" / "e" / "f" /
    "A" / "B" / "C" / "D" / "E" / "F"
```

```go
v := validator.New()
err := v.Var("479992BD-5A1E-43AA-9268-597F581380F0", "uuid4") // err will not be nil
```

alexbilbie:
Fixes #402.

Change Details:

- Added uppercase letters to regexs
- Added additional test case

@go-playground/admins


joeybloggs:
@alexbilbie thanks for the PR!

The change makes perfect sense, however this would also be a breaking change.

If it were me I would ensure the values are lowercase before validating using [mold](https://github.com/go-playground/mold/blob/v2/modifiers/modifiers.go) or something.

but I can see how both would be useful, instead of modifying the existing ones what do you think about adding new ones with the suffix `lax` eg. `uuidlax`?

that way both are available without a breaking change.

rssathe:
What's the status on this @joeybloggs @alexbilbie ? Understood it's a breaking change but it seems like the validator doesn't comply with the RFC, so I'm not sure of the usefulness of having both.

rssathe:
@joeybloggs can we get this merged in?

joeybloggs:
@rssathe It's breaking in the fact that people may be relying on it to catch is it's not lowercase, I **WILL NOT** make a breaking change to people like that.

I suggested a way to get it merged in and never heard back, @rssathe if you want to modify the code by adding the new validations with a new key eg. `uuid_rfc4122` I'd be more than happy to merge.

anyone who wants to still use `uuid` alias should work

rssathe:
It seems fairly deceptive to suggest something is validating a uuid when as according to the spec its actually not correctly doing so. In fact i'd suspect that most people have unexpected issues debugging what seemingly should be and looks like a valid UUID and is failing validation with this validator. My opinion would actually be the reverse if for some more peculiar reason you wanted to only validate lowercase uuids to have a validator called uuid_lowercase. But seeing as you have a strong opinion on this regarding a "breaking change" I can submit a pr with your suggestion.

joeybloggs:
@rssathe the trickey part is this statement
> RFC 4122 allows uppercase characters for input of functions but requires lowercase characters on output:

if it allowed in and out I'd gladly make the change, but unfortunately it doesn't. 

The idea behind the way it is now is that **usually** your validating before saving or persisting somewhere or forwarding somewhere else, so if it were me writing a program, to ensure that when it's output will always be correct, is lowercase before validation and what I encourage people to do to save any headaches.

rssathe:
@joeybloggs have a pr up for this!

alexbilbie:
Closing as it's fixed in #437

PR #415, opened 2018-11-09
Make it compatible with some mock library such as go-sqlmock

hxmhlt:
Change Details:
When I use some mock library such as go-sqlmock.It will expect some db operations in order.
But when I mock my project include some my own validator with validator.v8(custom validator tag) , some error occured.

Because code in here is range the struct field map and you know map in golang is out-of-order , I can not make it compatible with the mock sql in-order.

**So I change the  range of struct field map to range sorted mapKey slice and it solved.**

@go-playground/admins

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -549,7 +550,17 @@ func (v *Validate) tranverseStruct(topStruct reflect.Value, currentStruct reflec
 	// if present
 	if first || ct == nil || ct.typeof != typeStructOnly {
 
-		for _, f := range cs.fields {
+		//By hxmhlt @2018.11.9
+		// Modify range map to range sorted mapKey slice for sorting the struct field , promise my custom validation call ordered to make it compatible with some mock library such as gp-sqlmock
+		var keys []int
+		for k := range cs.fields {
+			keys = append(keys, k)
+		}
+		sort.Ints(keys)
+
+		for _, k := range keys {
+
+			f := cs.fields[k]
 
 			if partial {
 
--- a/validator.go
+++ b/validator.go
@@ -551,7 +551,7 @@ func (v *Validate) tranverseStruct(topStruct reflect.Value, currentStruct reflec
 	if first || ct == nil || ct.typeof != typeStructOnly {
 
 		//By hxmhlt @2018.11.9
-		// Modify range map to range sorted mapKey slice for sorting the struct field , promise my custom validation call ordered to make it compatible with some mock library such as gp-sqlmock
+		// Modify range map to range sorted mapKey slice for sorting the struct field , promise my custom validation call ordered to make it compatible with some mock library such as go-sqlmock
 		var keys []int
 		for k := range cs.fields {
 			keys = append(keys, k)


joeybloggs:
Hello @hxmhlt 

I'm not sure this will solve your problem as the resulting [ValidationErros](https://github.com/go-playground/validator/blob/5a5ca89ed7a584419499c2d974bff0d79916c97f/validator.go#L188) is also a map, and if multiple errors are returned, which this library supports.


I'm also worried about the performance impacts of this as ranging, balding and array, sorting and then ringing again with a map lookup isn't a cheap operations.

I curious if migrating to v9 wouldn't be a better option as the resulting errors are now an [array](https://github.com/go-playground/validator/blob/v9/errors.go#L37) rather than a map which should solve your issue; furthermore v9 is far more efficient and the interface vastly simplified.

do you think v9 would work for you better?

hxmhlt:
@joeybloggs Yes,it's work for me. Go-sqlmock is only incompatible with validatorv9 in this respect, because it requires that all sql execution order must be certain. Now the code here makes the execution order of the validator undefined.So I have to modify here to slice. Or  should you consider allowing the user to specify the order of execution of the validators in different fields?

joeybloggs:
@hxmhlt oh so you're saying v9 doesn't work with sqlmock also? would you be able to provide an example?

hxmhlt:
emmmm...
ex,I have a struct
`
type MemberRoleDTO struct {
	RoleId 		int		`json:"roleId" binding:"required,validateroleid"`
	MemberId	int		`json:"memberId" binding:"required,validatememberid"`
	MemberType  string	`json:"memberType" binding:"required,validatemembertype"`
	SourceId    int     `json:"sourceId" binding:""`
	SourceType  string	`json:"sourceType" binding:""`
}
`


`
func validateroleid(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string)bool{
	if targetField, ok := field.Interface().(int); ok {
		if _,err := role.FindOne(&role.Role{Id:targetField});err==nil{
			return true
		} else {
			return false
		}
	}
	return false
}

func validatememberid(v *validator.Validate, topStruct reflect.Value, currentStructOrField reflect.Value, field reflect.Value, fieldType reflect.Type, fieldKind reflect.Kind, param string)bool{
	if targetField, ok := field.Interface().(int); ok {


		if currentStructOrField.Interface().(*MemberRoleDTO).MemberType == "user"{
			if _,err :=users.FindOneUser(&users.User{Id:int64(targetField)});err==nil{
				return true
			}
		}else if currentStructOrField.Interface().(*MemberRoleDTO).MemberType  == "client"{
			if _,err :=auth.FindOneClient(&auth.Client{Tid:targetField});err==nil{
				return true
			}
		}else{
			return false
		}
	}
	return false
}
`
And validateroleid,validatememberid  is my validater func,In this 2 func I have to do "FindOne" to access my db, so I wrote my mock:
`	mock.ExpectQuery(`SELECT (.+) FROM "t_iam_role"`).WithArgs(1).
				WillReturnRows(sqlmock.NewRows([]string{"id", "name", "code", "description","level","is_enabled"}).
					FromCSVString("1,superAdmin,sysadmin,,user,1"))

			columns := []string{"id", "user_account","user_name","user_email","user_phone","dept","pwd"}
			mock.ExpectQuery(`SELECT (.+) FROM "t_user"`).WithArgs(1).
				WillReturnRows(sqlmock.NewRows(columns).FromCSVString(
					`1,admin,admin,admin@ctrip.com,110,,$2a$10$bCoOHdNZpLh3mIvzs19hJOOaBfegwosff5iOouyO7rhfDRy6s2aaW`))`

So the error is validateroleid and validatememberid func execute disordered.I can't sure the correct mock is written on the above, or is 
`
]columns := []string{"id", "user_account","user_name","user_email","user_phone","dept","pwd"}
			mock.ExpectQuery(`SELECT (.+) FROM "t_user"`).WithArgs(1).
				WillReturnRows(sqlmock.NewRows(columns).FromCSVString(
					`1,admin,admin,admin@ctrip.com,110,,$2a$10$bCoOHdNZpLh3mIvzs19hJOOaBfegwosff5iOouyO7rhfDRy6s2aaW`))

mock.ExpectQuery(`SELECT (.+) FROM "t_iam_role"`).WithArgs(1).
				WillReturnRows(sqlmock.NewRows([]string{"id", "name", "code", "description","level","is_enabled"}).
					FromCSVString("1,superAdmin,sysadmin,,user,1"))
`

joeybloggs:
Hey @hxmhlt been mulling this over quite a bit and going to mark this as `Wont Fix`

`v8` is really only in maintenance mode and the issue appears confined to v8. I do recognize this could still happen in `v9` merely because of the order of the fields, but in this case, once the validation complexity is this high, I really recommend doing a `Struct Level` validation also where you can tightly control the validation and its order.

Thank you again for the PR.

hxmhlt:
@joeybloggs OK,Thanks for your suggestion.

PR #420, opened 2018-11-16
Allow ValidationErrors type to be used externally

joshhsoj1902:
The `ValidationErrors` type is public, but inside its 2 functions it casts to the private `fieldError`. Since there is no way to access the `fieldError` outside this package it meant creating and using `ValidationErrors` outside this package was impossible.

My use case is for testing, I'm trying to test how my code behaves when various `ValidationErrors` are returned without needing to actually setup data to fail in the various ways.

Fixes Or Enhances # .

Change Details:

- 
- 
- 

@go-playground/admins

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -44,11 +44,11 @@ func (ve ValidationErrors) Error() string {
 
 	buff := bytes.NewBufferString("")
 
-	var fe *fieldError
+	var fe FieldError
 
 	for i := 0; i < len(ve); i++ {
 
-		fe = ve[i].(*fieldError)
+		fe = ve[i].(FieldError)
 		buff.WriteString(fe.Error())
 		buff.WriteString("\n")
 	}
@@ -61,10 +61,10 @@ func (ve ValidationErrors) Translate(ut ut.Translator) ValidationErrorsTranslati
 
 	trans := make(ValidationErrorsTranslations)
 
-	var fe *fieldError
+	var fe FieldError
 
 	for i := 0; i < len(ve); i++ {
-		fe = ve[i].(*fieldError)
+		fe = ve[i].(FieldError)
 
 		// // in case an Anonymous struct was used, ensure that the key
 		// // would be 'Username' instead of ".Username"
@@ -155,6 +155,9 @@ type FieldError interface {
 	// NOTE: is not registered translation can be found it returns the same
 	// as calling fe.Error()
 	Translate(ut ut.Translator) string
+
+	// Error returns the FieldError's error message
+	Error() string
 }
 
 // compile time interface checks
--- a/errors.go
+++ b/errors.go
@@ -61,10 +61,10 @@ func (ve ValidationErrors) Translate(ut ut.Translator) ValidationErrorsTranslati
 
 	trans := make(ValidationErrorsTranslations)
 
-	var fe FieldError
+	var fe *fieldError
 
 	for i := 0; i < len(ve); i++ {
-		fe = ve[i].(FieldError)
+		fe = ve[i].(*fieldError)
 
 		// // in case an Anonymous struct was used, ensure that the key
 		// // would be 'Username' instead of ".Username"


errors.go
```diff
@@ -155,6 +155,9 @@ type FieldError interface {
 	// NOTE: is not registered translation can be found it returns the same
 	// as calling fe.Error()
 	Translate(ut ut.Translator) string
+
+	// Error returns the FieldError's error message
+	Error() string
```
joeybloggs: Hey @joshhsoj1902 thanks for the PR

I think that only this addition is necessary

joshhsoj1902: Hey @joeybloggs, I might be doing something wrong on my end, But for me, it only works when I also change the `*fieldError` references to the interface. When I just add `Error() string` to the interface I still get this error thrown

>panic: interface conversion: validator.FieldError is *validator.fakeFieldError, not *validator.fieldError 

(that panic happens on line 51, `fakeFieldError` is my implentation of `FieldError`)

My understanding is that since line 51 is trying to cast to the private struct, the only way the casting will pass is if the object being cast was also made from that private struct. and since it's private, I have no way of actually doing that.

joeybloggs: I'll take a closer look, sorry for the delay I'm extremely busy

joshhsoj1902: No problem at all. If there is anything you would rather I do instead, let me know. I think there are a few ways to solve this.

joeybloggs: @joshhsoj1902 would you be able to share your example, seems like the best place for me to start :)

joshhsoj1902: Hey Joey, sorry for the slow reply. I do plan on circling back and creating a small sample program for you. Time is always the issue. Things should slow down a bit for me after next week.

joeybloggs: no worries I only get a couple hours at most to take care of my open source stuff lately, tis the holiday season.

joeybloggs:
Hey @joshhsoj1902 I would be very interested in seeing your example that fails, I was able to cast the error to the field error type, see modified simple example below; strangely enough, I was unable to cast to the interface type in the range itself, had to be after the variable was already extracted.
```go
package main

import (
	"fmt"
	"reflect"

)

// User contains user information
type User struct {
	FirstName      string     `validate:"required"`
	LastName       string     `validate:"required"`
	Age            uint8      `validate:"gte=0,lte=130"`
	Email          string     `validate:"required,email"`
	FavouriteColor string     `validate:"iscolor"`                // alias for 'hexcolor|rgb|rgba|hsl|hsla'
	Addresses      []*Address `validate:"required,dive,required"` // a person can have a home and cottage...
}

// Address houses a users address information
type Address struct {
	Street string `validate:"required"`
	City   string `validate:"required"`
	Planet string `validate:"required"`
	Phone  string `validate:"required"`
}

// use a single instance of Validate, it caches struct info

func main() {

	validate = validator.New()

	validateStruct()
	validateVariable()
}

func validateStruct() {

	address := &Address{
		Street: "Eavesdown Docks",
		Planet: "Persphone",
		Phone:  "none",
	}

	user := &User{
		FirstName:      "Badger",
		LastName:       "Smith",
		Age:            135,
		Email:          "Badger.Smith@gmail.com",
		FavouriteColor: "#000-",
		Addresses:      []*Address{address},
	}

	// returns nil or ValidationErrors ( []FieldError )
	err := validate.Struct(user)
	if err != nil {

		if _, ok := err.(*validator.InvalidValidationError); ok {
			fmt.Println(err)
			return
		}

			e2 := err.(validator.FieldError)
			fmt.Println(reflect.TypeOf(e2))
			fmt.Println(e2.Namespace())
			fmt.Println(e2.Field())
			fmt.Println(e2.StructNamespace()) // can differ when a custom TagNameFunc is registered or
			fmt.Println(e2.StructField())     // by passing alt name to ReportError like below
			fmt.Println(e2.Tag())
			fmt.Println(e2.ActualTag())
			fmt.Println(e2.Kind())
			fmt.Println(e2.Type())
			fmt.Println(e2.Value())
			fmt.Println(e2.Param())
			fmt.Println()
		}

		// from here you can create your own error messages in whatever language you wish
		return
	}

	// save user to database
}

func validateVariable() {

	myEmail := "joeybloggs.gmail.com"

	errs := validate.Var(myEmail, "required,email")

	if errs != nil {
		fmt.Println(errs) // output: Key: "" Error:Field validation for "" failed on the "email" tag
		return
	}

	// email ok, move on
}
```

deankarn:
@joshhsoj1902 are you able to provide an example?

joshhsoj1902:
Hey @deankarn, Sorry for never getting back to you on this, I ended up changing projects and the code I was dealing with I no longer have access to. I would need to come up with a new example and honestly, I don't foresee myself having time to do that anytime soon. 

I'm really sorry, I hate not being able to help out more :(.

Feel free to close this.

deankarn:
I you ever do find time or want to pursue again please don't hesitate to revive this @joshhsoj1902 :)

DavidLarsKetch:
I've also run into the desire to include the `fieldError` message when handling and passing along error messages from `ValidationErrors`; so I've opened a new PR with that small change.

PR #421, opened 2018-11-17
Add NotBlank validation function.

andreiavrammsd:
Change Details:

- There are cases when a value needs to not be empty (like a string containing only spaces, slice not nil but with no elements) and I added a new tag to easily overcome this need.

```
type TestString struct {
	NotBlank  string `validate:"notblank"`
}
```

@go-playground/admins

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -245,6 +245,14 @@ ensures the value is not nil.
 
 	Usage: required
 
+NotBlank
+
+This validates that the value is not blank or with length zero.
+For strings ensures they do not contain only spaces. For channels, maps, slices and arrays
+ensures they don't have zero length. For others, the "required" validation is used.
+
+	Usage: notblank
+
 Is Default
 
 This validates that the value is the default value and is almost the
--- a/doc.go
+++ b/doc.go
@@ -245,14 +245,6 @@ ensures the value is not nil.
 
 	Usage: required
 
-NotBlank
-
-This validates that the value is not blank or with length zero.
-For strings ensures they do not contain only spaces. For channels, maps, slices and arrays
-ensures they don't have zero length. For others, the "required" validation is used.
-
-	Usage: notblank
-
 Is Default
 
 This validates that the value is the default value and is almost the
@@ -998,5 +990,25 @@ that should not make it to production.
 	}
 
 	validate.Struct(t) // this will panic
+
+Non standard validators
+
+	type Test struct {
+		TestField string `validate:"yourtag"`
+	}
+
+	t := &Test{
+		TestField: "Test"
+	}
+
+	validate := validator.New()
+	validate.RegisterValidation("yourtag", validations.ValidatorName)
+
+	NotBlank
+		This validates that the value is not blank or with length zero.
+		For strings ensures they do not contain only spaces. For channels, maps, slices and arrays
+		ensures they don't have zero length. For others, a non empty value is required.
+
+		Usage: notblank
 */
 package validator
--- a/doc.go
+++ b/doc.go
@@ -993,6 +993,11 @@ that should not make it to production.
 
 Non standard validators
 
+A collection of validation rules that are frequently needed but are more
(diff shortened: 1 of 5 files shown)


joeybloggs:
Hey @andreiavrammsd sorry for the delay I responded a few days ago but I don't see it here, was having internet issues that day...anyways

I'm curious how this is different from combining `required` and `gt` like so `required,gt=0`?
if it's just that you want the `nonblank` tag as the error an alias can be registered for this.

andreiavrammsd:
Hello,

The value "" (empty string) is caught by required,gt=0, but "     " (string made only of spaces) is not. In many situations, I need to make sure there is actually a value inside a string, not only spaces.
I'm using a custom registered tag now, but I see the situation as quite common.

Thanks!

joeybloggs:
Ah I see, I missed the TrimSpace in your validation; I'm not against this but also wanted to suggest an alternate approach; I too have seen that things like this are pretty common and so created another package to help sanitize the data prior to validation.

[mold](https://github.com/go-playground/mold) which will allow you to trim and sanitize data in a number of ways, even trimming values of maps.

My only concern about adding this to the validation lib is that I can see a great number of additional cases eg. streaming map string values or map of map string values... and don;t think it would make sense to have all of them in the std validation package.

What do you think about using the [modifiers](https://github.com/go-playground/mold/blob/v2/modifiers/modifiers.go) package of the mold lib prior to validation?

andreiavrammsd:
Regarding trimming, is it worth using a library for a simple built-in function?
For maps and slices, you could use "dive" in order to run the validation on everything.

andreiavrammsd:
If it feels to heavy for the standard library, no problem, we can close it.
It's just something I see as a standard for validations.

joeybloggs:
Sorry for the delay, super busy, my only concern is that the validation actually does more that one thing that corresponds https://github.com/go-playground/validator/pull/421/files#diff-e8dcd0f6e62fb5a9257bc99c0891f3f8R1197

eg. both `len` and `required` are being mixed together in this validation; the way I look at baked-in validations is that they are of singular purpose.

but I definitely see the value in this and other non-standard validations as well so I have a proposal, if we could create some sub-folders for this package `non-standard/validators` and `non-standard/aliases`

Then we can add these sorts of validators here and people can easily register them if they desire, but won't be registered by default.

What do you think?

andreiavrammsd:
This could be something. How would you see the non-standard validators registration: (1) including tags or (2) leaving the user the choose a custom tag?

1. New register function: this would automatically register the validate function with the tag "notblank", so you'd use it as:
```
type User struct {
    Username `validate:"notblank"`
}
v := validator.New()
v.RegisterNonStd(validators.NotBlank) // please ignore the "RegisterNonStd" name, I'm just out of ideas for now
```

2. Allow user to use the function with any tag:
```
type User struct {
    Username `validate:"customtag"`
}
v := validator.New()
v.RegisterValidation("customtag", validators.NotBlank)
```

joeybloggs:
@andreiavrammsd I like option #2 but I'd still want them in another package as described because there are still aliases and translations that can go along with the validation and I'm worried about cluttering up the `validator` namespace.

andreiavrammsd:
Is it starting to look as you thought?

joeybloggs:
@andreiavrammsd apologies for the delay, trying to catch up after the holidays, I have not forgotten

doc.go
```diff
@@ -990,5 +990,25 @@ that should not make it to production.
 	}
 
 	validate.Struct(t) // this will panic
+
+Non standard validators
```
joeybloggs: If you could add a note here about what Non standard validations are and that you have to register them manually that would be amazing :)

joeybloggs:
@andreiavrammsd looking great! if you could add a note about non-standard then I can merge :)

joeybloggs:
Awesome thanks!

PR #426, opened 2018-11-28
Make repo a Go module

mastertinner:
Fixes Or Enhances

Change Details:

- Add a `go.mod` and a `go.sum` file that describe the dependencies and make this repo a Go module.

@go-playground/admins

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,7 @@
+module gopkg.in/go-playground/validator.v9
+
+require (
+	github.com/go-playground/locales v0.12.1
+	github.com/go-playground/universal-translator v0.16.0
+	gopkg.in/go-playground/assert.v1 v1.2.1
+)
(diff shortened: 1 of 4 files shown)


mastertinner:
Another discussion should be whether to move from importing from `gopkg.in` to directly importing from `github.com`.

joeybloggs:
Hey @mastertinner thanks,

I was considering this and once go modules are on by default in the next release I was going to convert this library and probably change stop using gopkg.in; it will require a version bump as it's a breaking change, but I have a few other things that also require breaking changes.

I think the changes in this PR have already been fixed by another PR.

wcsiu:
Can someone please add the go.mod, go.sum? It shouldn't break anything and the library will be go module ready.

offlinehacker:
Same issue here, i guess this is the only decent validation library in go, and can't even use it without forking

deankarn:
🤣 ok, guys, some of the comments here are quite humorous and outlandish.

Now that Go modules has been released and because this library is > v1 I now know that migrating **IS** a breaking change now that the versions affect the import paths within other programs(totally crazy).

I will start on a v10 branch when I can find some time to avoid any breaking changes and other headaches and issues that Go modules has and has brought to the table.

deankarn:
****

PR #437, opened 2019-01-10
Fixes #402 UUID RFC4122 Validation

### LINKED ISSUE #402: Incorrect UUID validation regex

rssathe:
Fixes Or Enhances #402  .

Change Details:

- 
- 
- 

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
-	urn "github.com/leodido/go-urn"
 	"net"
 	"net/url"
 	"os"
@@ -15,6 +14,8 @@ import (
 	"sync"
 	"time"
 	"unicode/utf8"
+
+	urn "github.com/leodido/go-urn"
 )
 
 // Func accepts a FieldLevel interface for all validation needs. The return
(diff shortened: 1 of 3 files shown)


rssathe:
@joeybloggs any chance you can look at this?

joeybloggs:
@rssathe LGTM, I'll merge now, it would be great if you could add these to the docs also :)

PR #440, opened 2019-01-24
Allow latitude/longitude validation for numeric values

RaMin0:

Change Details:
- Make latitude/longitude validation work with types other than `string`s.

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -304,12 +304,12 @@ func isSSN(fl FieldLevel) bool {
 
 // IsLongitude is the validation function for validating if the field's value is a valid longitude coordinate.
 func isLongitude(fl FieldLevel) bool {
-	return longitudeRegex.MatchString(fl.Field().String())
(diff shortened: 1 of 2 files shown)


baked_in.go
```diff
@@ -304,12 +304,12 @@ func isSSN(fl FieldLevel) bool {
 
 // IsLongitude is the validation function for validating if the field's value is a valid longitude coordinate.
 func isLongitude(fl FieldLevel) bool {
-	return longitudeRegex.MatchString(fl.Field().String())
+	return longitudeRegex.MatchString(fmt.Sprint(fl.Field().Interface()))
```
joeybloggs: like the change, it would be nice if you could add a type switch as it will allow the conversion from int to string to be much more efficient eg. https://github.com/go-playground/validator/blob/52ea44899849049e3ae9267ef1902c36807fe198/baked_in.go#L193

RaMin0: Did a small benchmark to compare and you're right. Check the latest commits.

PR #463, opened 2019-03-27
Fix email regex

benpaxton-hf:

Change Details:

- Unescape a `(?:` in the email regex that shouldn't be escaped
- Add a `)` in the correct position to close the above
- Add additional tests to verify correctness of email regex (both new tests fail with the old regex)

GitHub seems to not show the diff very nicely; the previously-escaped `(?:` is just before the second instance of `x01` in the regex; the new `)` is just before the `@`. `git show <commit> --word-diff=color --word-diff-regex=.` may help view this diff properly.

@go-playground/admins


joeybloggs:
Thanks @benpaxton-hf for the PR, will try to look at it this week

PR #469, opened 2019-04-29
Add `required_with`, `required_with_all`, `required_without`, `required_without_all` validators

mashmooli:

Change Details:
add these validators
- **required_with**: The field under validation must be present and not empty only if any of the other specified fields are not empty, (required_with:field1,field2);
- **required_with_all**: The field under validation must be present and not empty only if all of the other specified fields are not empty, (required_with_all:field1,field2);
- **required_without**: The field under validation must be present and not empty only when any of the other specified fields are empty, (required_without:field1,field2);
- **required_without_all**: The field under validation must be present and not empty only when all of the other specified fields are empty, (required_without_all:field1,field2).

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -62,108 +62,109 @@ var (
 	// you can add, remove or even replace items to suite your needs,
 	// or even disregard and use your own map if so desired.
 	bakedInValidators = map[string]Func{
-		"required":          hasValue,
-		"required_with":     requiredWith,
-		"required_with_all": requiredWithAll,
-		"required_without":  requiredWithout,
-		"isdefault":         isDefault,
-		"len":               hasLengthOf,
-		"min":               hasMinOf,
-		"max":               hasMaxOf,
-		"eq":                isEq,
-		"ne":                isNe,
-		"lt":                isLt,
-		"lte":               isLte,
-		"gt":                isGt,
-		"gte":               isGte,
-		"eqfield":           isEqField,
-		"eqcsfield":         isEqCrossStructField,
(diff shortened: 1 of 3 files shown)


deankarn:
Hey @mashmooli thanks for the PR! nice work!

I have a couple of requests before I can merge:
- can you add some more test cases? it looks like the current ones don't test all of the new code paths yet.
- can you add these new methods to the docs [here](https://github.com/go-playground/validator/blob/v9/doc.go), I'm sure a ton of other people will find the new validations useful as well :)

mashmooli:
Hi @deankarn thanks for the reply.
I added new methods to the docs and I'll be working on test cases soon.

deankarn:
@mashmooli merged, I think that a cool addition would be allowing this to work with nested structures also :)

The helper functions exist within this repo if you're interested :) 
eg. https://github.com/go-playground/validator/blob/v9/util.go#L60

andig:
@mashmooli hope it is ok to be asking here. I had  expected the following to throw an error but it isn't:

	func TestModbusConfigFoo(t *testing.T) {
		type Settings struct {
			Opt, Val string
		}

		cc := struct {
			Model    string
			Value    string
			Settings *Settings `validate:"required_without_all=Model Value"`
		}{
			Model: "foo",
			// Value: "bar",
		}

		validate := validator.New()
		if err := validate.Struct(cc); err == nil {
			t.Error(err)
		}
	}

Shouldn't that be covered by `required_without_all`?

PR #497, opened 2019-08-02
Add co dependent validations

pantsmann:
Enhancements for v9 that added support for codependent field validations such as for mutually exclusive fields.

I'm brand new to Go so any and all suggestions and comments will be appreciated. 

In particular, I'm uncomfortable with the function AddGroupField() being passed an interface{} when it expects it to be a *validate object. The passed object should be the FieldLevel object passed to validation functions but depends on the validate struct which isn't available externally. Perhaps there is a better way of doing this?

Change Details:

 * Added validationFlag's to enable special treatment for custom validations (and baked-ins). These flags can be passed in when registering custom validations (or through bakedInFlags). In particular, codependent field type validations may need one or more of these flags to run properly. I defined three flags: 
   * VFlagRunOnNil - run the validation even if the field is a nil ptr or interface; 
   * VFlagRunOnStruct - run validation at the struct level (only required, omitempty, and struct level functions run at this level otherwise); 
   * VFlagCoDependentErr - error should output in a different format for codependent field validations. 
 * Added CoDependentGroups global variable to store/manage named groups of codependent fields. This variable and its functions are exported to allow custom codependent validations. Functions include:
   * AddGroupField() - assigns a field to a named group.
   * CloseGroup() - marks a group as validated (prevents adding fields to a group after it has already been validated).
   * ClearGroups() - removes group(s) (all groups by default or just those passed in). validator.New() calls this function, but it can be called manually if needed.
   * Count() - returns the number of groups defined.
   * Fields() - returns the slice of fields (validate objects that can be passed to Func functions) in the given group.
   * FieldsOk() - same as above but also returns ok to specify if the given group exists.
 * Updated validator code to support codependent validations:
   * Set certain variables prior to validation func call in order to make them available validation funcs.
   * Added code to run flagged validations even if the field is a nil pointer of interface.
   * Added code to run flagged validations at the struct level  (for codependent fields that are sub-structs). Note that groups validations never fail so validations on sub-struct fields will also run, but when at_least, no_more_than, and mutex fail, validations do not run on sub-struct fields.
 * Added 'kind' field to cField struct to identify the original base kind of the field (not the internal type used in most cases)
 * Added four baked in validation functions that run common basic codependent field validations. For purposes of these functions 'non-empty' is defined using required with two additional requirements: (a) interface{} values are evaluated as their runtime type not as interface{} and (b) slices, arrays, and maps must have a non-zero length. Functions are the following:
   * group - Takes a group name and adds the field to that group. This function doesn't actually validate anything so it always returns true.
   * at_least - Takes a group name and an optional int N. Verifies that at least N fields in the group are non-empty. N defaults to 1 if not included. Closes group so it should be specified only on the last field in the group. Further codependent validations may be run on the group, but no new fields may be added to it after running this validation. 'group' does not need to be specified on the field where this validation is specified as this function also adds the field to the group prior to testing.
   * no_more_than - Same as at_least except that is verifies that no more than N fields in the group are non-empty.
   * mutex - Takes a group name only and verifies that no more and no less than one field in a group is non-empty. Otherwise, it functions the same as at_least.
 * Added additional format for error output string so that all fields in a codependency group will be listed. Custom validations can access this format by passing the VFlagCoDependentErr flag on registration.
 * Updated doc.go
 * Updated tests with codependent validation tests (Noticed that some existing tests appear to fail. I verified that this was also the case before I made my changes)
 * **_examples/custom-codependent/main.go** - Added example of a custom validation implementing a codependent fields check. This check ensures that the sum of three fields is greater than or equal to a given value. The important points: 
   * when registering the custom validation a validation flag is passed in that changes the default error message for the validation, 
   * the validation function accesses the CoDependentGroups global variable in order to gain access to and manage the codependent fields.

@go-playground/admins


pantsmann:
Working on adding default English translations for at_least, no_more_than, and mutex.

pantsmann:
Note: I realize that some of the functionality in this pull-request could already be done with required_without and required_without_all. The benefit of this new method is that it only runs the validation check once and only produces a single error upon failure. Reproducing mutex with required_without_all on three fields would run the validation check three times and produce three errors (one for each field). 

Additionally, this pull request adds greater flexibility for custom validations through validation flags and CoDependentGroups structure.

deankarn:
@pantsmann thanks for the PR!

It's a lot to take in and don't want to rush integrating something like this, please be patient.

pantsmann:
@deankarn no worries. There are still a few translation-related changes I'd like to make but I haven't been able to get around to them for a few weeks.

deankarn:
Closing due to age this really requires a new major version as it's a lot of changes.

PR #510, opened 2019-09-03
Perform validations on nil pointers

zenovich:
Fixes Or Enhances # .

Change Details:

Currently for pointer-type fields with nil values all the validations are skipped and an error mentioning to the first validation on the list is added. But what if we want to consider nil-values valid? For example, in our program we have a list of fields provided by user and, when we validate a form, we want to add errors for missing required fields, but not for nil fields that are given.

This PR allows to make fields with nil values valid and perform validations on them.

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -185,15 +185,27 @@ func parseOneOfParam2(s string) []string {
 }
 
 func isURLEncoded(fl FieldLevel) bool {
-	return uRLEncodedRegex.MatchString(fl.Field().String())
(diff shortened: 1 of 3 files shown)


deankarn:
This can now be done by registering a validation which allows nil values to run the function since https://github.com/go-playground/validator/releases/tag/v9.30.0

see https://github.com/go-playground/validator/blob/v9/validator_test.go#L8843

PR #511, opened 2019-09-03
Let validators see current fields paths

zenovich:
Fixes Or Enhances # .

Change Details:

Currently there is no way for validators to know which fields they are checking. So validators depending not only on the value, but on the field itself (e.g. checking that the field is listed in some external source) cannot work.

This PR adds new methods Path() & StructPath() into the FieldLevel interface allowing validators to get a path (based on FieldName-s) or a struct path (based on StructFieldName-s) of the current field respectively.

@go-playground/admins

### DIFF
--- a/field_level.go
+++ b/field_level.go
@@ -23,6 +23,13 @@ type FieldLevel interface {
 	// returns the struct field's name
 	StructFieldName() string
 
+	// returns the field's path with tag names
+	// taking precedence over fields actual names
+	Path() string
+
+	// returns the struct field's path
+	StructPath() string
+
 	// returns param for validation against current field
 	Param() string
 
@@ -56,6 +63,17 @@ func (v *validate) FieldName() string {
 // StructFieldName returns the struct field's name
 func (v *validate) StructFieldName() string {
 	return v.cf.name
+}
+
+// Path returns the field's path with tag names
+// taking precedence over fields actual names.
+func (v *validate) Path() string {
+	return v.str1
+}
+
+// StructPath returns the struct field's path
+func (v *validate) StructPath() string {
+	return v.str2
 }
 
 // Param returns param for validation against current field
--- a/validator.go
+++ b/validator.go
@@ -175,15 +175,15 @@ func (v *validate) traverseField(ctx context.Context, parent reflect.Value, curr
 					v.cf = cf
 					v.ct = ct
 
+					v.str1 = string(append(ns, cf.altName...))
+
+					if v.v.hasTagNameFunc {
+						v.str2 = string(append(structNs, cf.name...))
+					} else {
+						v.str2 = v.str1
+					}
+
 					if !ct.fn(ctx, v) {
-						v.str1 = string(append(ns, cf.altName...))
-
-						if v.v.hasTagNameFunc {
-							v.str2 = string(append(structNs, cf.name...))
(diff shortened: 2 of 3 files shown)


deankarn:
Hey @zenovich can you elaborate on why having these paths is useful within the validators?

I'm concerned that the code has been moved outside of the validation check which means that this logic and allocations will now happen for every field being validated, even if there's no validation error. I'm inclined not to accept the PR unless there's a very good reason.

zenovich:
Hi @deankarn 
The problem is that we need these paths during the validation process itself, not for errors. We decide if the field is valid basing on the paths. This means that this logic really should happen for every field being validated, otherwise we cannot even decide if there's an error or not.

deankarn:
@zenovich could you provide a real world example of when these paths would be used? I can't think of any off the top of my head.

zenovich:
@deankarn Consider we have a structure with data:struct{ a struct {b *string}; b *string} (it is valid for any of "b"-s to be nil). Also, we have a map filled on parsing a web-request, letting us know which of the fields are given by the user: paramsGiven = map[string]bool{"a":true, "b": true, "a.b": false}. Now imagine that we need to validate "b" & "a.b" in the same way, but only if they are given by the user (remember that nil is a valid value and we cannot consider it as "not given"). To solve this problem, we use a custom validator which searches for a field's path in "paramsGiven" before doing real checks. That's why we need to be able to get correct fields path in validators.

deankarn:
@zenovich I'm wondering why the current functionality would not work for you.
you are able to pass external information through the [context.Context](https://github.com/go-playground/validator/blob/5b97da4cbf655ab460ec928f6737baa6edd2e74d/baked_in.go#L27) to the validation and struct validation currently and there are [built-in helpers for traversing strings by namespace/path](https://github.com/go-playground/validator/blob/fb6c45823ad5db49e788621271c6f86fa3a0f149/field_level.go#L58) just like in your example.

zenovich:
> @zenovich I'm wondering why the current functionality would not work for you.
> you are able to pass external information through the [context.Context](https://github.com/go-playground/validator/blob/5b97da4cbf655ab460ec928f6737baa6edd2e74d/baked_in.go#L27) to the validation and struct validation currently and there are [built-in helpers for traversing strings by namespace/path](https://github.com/go-playground/validator/blob/fb6c45823ad5db49e788621271c6f86fa3a0f149/field_level.go#L58) just like in your example.

These build-in helpers were added two months after this PR, so I don't know if they can help or not.

zenovich:
I'll check

zenovich:
@deankarn It looks like those new build-in helpers still don't allow a validator to figure out which of the fields it is checking at the moment. And without this information a validator cannot check if the field was given by the user or not (no matter whether this information is stored in the context or somewhere else).

deankarn:
@zenovich would [StructPartial](https://pkg.go.dev/github.com/go-playground/validator/v10#Validate.StructPartial) be what you're looking for to filter only the fields that need to be validated using the map example?

zenovich:
@deankarn, not really. Actually the StructPartial does the opposite to what I need. I need to validate all the fields, but my custom validator should know which field it is validating at the moment.

deankarn:
Sorry closing due to age. This is something I will think about for the next version.

PR #512, opened 2019-09-07
Optimize required_with_*

dchaofei:
Fixes Or Enhances # .

Change Details:
- End the loop of the requiredWith* method early, I am tempted to use break or goto, and finally I chose goto, which is more readable.

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -1350,9 +1350,11 @@ func requiredWithAll(fl FieldLevel) bool {
 
 		if !requireCheckFieldKind(fl, param) {
 			isValidateCurrentField = false
+			goto NEXT
 		}
 	}
 
+NEXT:
 	if isValidateCurrentField {
 		return requireCheckFieldKind(fl, "")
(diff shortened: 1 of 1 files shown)


leafduo:
`goto` is generally considered harmful, and `goto` in golang has limitations like cannot jump over variable declarations. `break` is completely fine here.

dchaofei:
@leafduo But go-playground/validator uses a lot of goto, so I want to be consistent

deankarn:
@dchaofei thanks you for the PR!

in this case it would be great if the `goto`'s could be replaced with `break`'s, although there are a number of `goto`'s within the codebase they are used in cases where:
- break isn't available or would not work
- to avoid breaking the program into too many small functions causing runtime performance overhead.

deankarn:
@dchaofei I optimized this with some other fixes in https://github.com/go-playground/validator/releases/tag/v9.30.0

closing :)

Issue #525, opened 2019-10-01
Validate oneof with values containing space


muhammadkholidb:
### Package version eg. v8, v9: 

v9

### Issue, Question or Enhancement:

Is it possible to use **oneof** validation with values containing space character? Tried with single quote but it did not seem to work.

### Code sample, to showcase or reproduce:

```go

type Data struct {
    Status string `json:"status" validate:"required,oneof='Awaiting Verification' 'Verified' 'Failed Verification'"`
}

```

deankarn:
@muhammadkholidb no, unfortunately, this isn't supported for `oneoff` at the moment.

As a workaround for now though is you can register your own custom validation function which accomplishes the same thing :)

JonathanWThom:
Hello 👋 , I'm going to take a crack at this.

fr3fou:
Isn't this closed by #541

PR #541, opened 2019-11-17
Adds ability to validate oneof for space separated strings

JonathanWThom:
Fixes Or Enhances https://github.com/go-playground/validator/issues/525.

Change Details:

* Adds the ability to match on space separated strings when using the
`oneof` validation. Space separted strings must be surrounded by single
quotes to be validated as one string. For example:

```
oneof='Awaiting Verification' 'Verified' 'Failed Verification'
```

passes validation for a field that is exactly `Failed Verification`
(though just `Failed` would...fail).

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -177,7 +178,8 @@ func parseOneOfParam2(s string) []string {
 	oneofValsCacheRWLock.RUnlock()
 	if !ok {
 		oneofValsCacheRWLock.Lock()
-		vals = strings.Fields(s)
+		re := regexp.MustCompile(`'[^']*'|\S+`)
+		vals = re.FindAllString(s, -1)
 		oneofValsCache[s] = vals
 		oneofValsCacheRWLock.Unlock()
 	}
@@ -213,7 +215,8 @@ func isOneOf(fl FieldLevel) bool {
 		panic(fmt.Sprintf("Bad field type %T", field.Interface()))
 	}
 	for i := 0; i < len(vals); i++ {
-		if vals[i] == v {
+		val := strings.Replace(vals[i], "'", "", -1)
+		if val == v {
 			return true
 		}
 	}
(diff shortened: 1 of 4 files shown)


baked_in.go
```diff
@@ -177,7 +178,8 @@ func parseOneOfParam2(s string) []string {
 	oneofValsCacheRWLock.RUnlock()
 	if !ok {
 		oneofValsCacheRWLock.Lock()
-		vals = strings.Fields(s)
+		re := regexp.MustCompile(`'[^']*'|\S+`)
```
deankarn: Can this be moved out out the hotpath? eg. a global variable 🙏

baked_in.go
```diff
@@ -213,7 +215,8 @@ func isOneOf(fl FieldLevel) bool {
 		panic(fmt.Sprintf("Bad field type %T", field.Interface()))
 	}
 	for i := 0; i < len(vals); i++ {
-		if vals[i] == v {
+		val := strings.Replace(vals[i], "'", "", -1)
```
deankarn: instead of separating this logic from the `parseOneOfParam2` function is it possible to do the replace before adding to the `oneofValsCache` cache, then the output of the function is constent and all other validations can benefit from the same changes 😄  eg. `required_with*`

baked_in.go
JonathanWThom: Good idea, I will do that.

baked_in.go
JonathanWThom: I'll give this a shot 👍

JonathanWThom:
Thanks for taking a look @deankarn. I'll try to find time to address your comments in the coming week.

JonathanWThom:
I believe I've accommodated your feedback - let me know if there's anything else that could be improved. Thanks!

JonathanWThom:
Thanks for merging this and https://github.com/go-playground/validator/pull/548 @deankarn!

deankarn:
Np I still have to cut a release, merging and porting some other things before cutting one also :)

lhphongclassfunc:
> Thanks for merging this and #548 @deankarn!

thanks for your contribution, today i encountered this case and solved it successfully 💪🏻

PR #578, opened 2020-02-10
Add list of baked in validations in README.md

### LINKED ISSUE #509: Document list of all available rules for increased productivity
v9

During development, it will be extremely useful to have a list of available validation methods available at one place for reference. I understand that I can look at the package to see the available validation methods, but switching back and forth between packages is slow.

I want to request a *.md file which I can open in a browser tab and quickly reference the validation to check if rule is called min or minimum for example.

An example to refer would be https://respect-validation.readthedocs.io/en/1.1/list-of-rules/
where I can see everything quickly. I think a list of this kind would really boost developer productivity.

```go
type RegisterRequest struct {
       // an md file will be useful to quickly see if rule is called alpha, alphanumeric or alphabet
	FirstName   string `json:"first_name" validate:"required,min=3,alpha"`
	LastName    string `json:"last_name"`
}
```

psampaz:
Fixes  #509 

@go-playground/admins

Todo list:

- [ ] Decide on groups
- [ ] Sort all tags alphabetically
- [ ] Update pull request template with checkbox for updating both README and doc.go in case of a new validation

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@ Package validator implements value validations for structs and individual fields
 
 It has the following **unique** features:
 
--   Cross Field and Cross Struct validations by using validation tags or custom validators.  
+-   Cross Field and Cross Struct validations by using validation tags or custom validators.
 -   Slice, Array and Map diving, which allows any or all levels of a multidimensional field to be validated.
--   Ability to dive into both map keys and values for validation  
+-   Ability to dive into both map keys and values for validation
 -   Handles type interface by determining it's underlying type prior to validation.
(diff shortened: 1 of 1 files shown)


psampaz:
@deankarn have a look. 

I've included the whole list of baked-in validations are seen here https://github.com/go-playground/validator/blob/master/baked_in.go#L65. 

Please arrange the groups as you wish.  I made an initial grouping just for the example.

PrakharSrivastav:
Thank you for all the efforts to assemble all validations at a single place.

Issue #599, opened 2020-04-19
RFC952 Hostname validation validate multiple consecutive dots. 


amiraliucsc:
### Package version eg. v8, v9: 
V9

### Issue, Question or Enhancement:
Current regex used for hostname_rfc952 will validate a host like the following:

example.........another.com

This is because the regex hostnameRegexStringRFC952 is having [a-zA-Z0-9\-\.]+ which means it can validate 1 or more dot ('.') character. 
To fix this a new regex has been used which only accepts a dot valid only if another character is before the dot. 

### Code sample, to showcase or reproduce:

```go
func validate_host(host string) bool {
    host = strings.TrimSpace(host)
    validate := validator.New()
    err := validate.Var(host, "hostname")
    if err != nil {
        return false
    }
    return true
}

is_valid :=  validate_host("example....................com")
fmt.Printf("%v", is_valid)  // true

```

deankarn:
PR Merged, closing :)

PR #600, opened 2020-04-19
Fix/rfc952 host

amiraliucsc:
Signed-off-by: amiraliucsc <amirali7089@gmail.com>

Fixes Or Enhances #599 .

Change Details:
This PR will change the hostnameRegexStringRFC952 to validate the dots only if one character is before the dot. This is to prevent considering string such as "example.....com" as a valid host. 

@go-playground/admins


amiraliucsc:
Any update on this fix?

Issue #597, opened 2020-04-13
How to translate the param field?


cncal:
### Package version: 

v10

### Enhancement:

Is it possible to translate the param field?

### Code sample:

```go
package main

import (
	"fmt"
	"reflect"

	"github.com/go-playground/locales/en"
	ut "github.com/go-playground/universal-translator"
	"github.com/go-playground/validator/v10"
	en_translations "github.com/go-playground/validator/v10/translations/en"
)

var (
	uni      *ut.UniversalTranslator
	validate *validator.Validate
)

func main() {
	en := en.New()
	uni = ut.New(en, en)
	trans, _ := uni.GetTranslator("en")

	validate = validator.New()
	en_translations.RegisterDefaultTranslations(validate, trans)

	validate.RegisterTagNameFunc(func(field reflect.StructField) string {
		locale := field.Tag.Get(`json`)
		if locale == `` || locale == `-` {
			return ``
		}
		return locale
	})

	type Num struct {
		Num1 uint8 `json:"num1"`
		Num2 uint8 `validate:"gtfield=Num1" json:"num2"`
	}

	num := Num{
		Num1: 0,
		Num2: 0,
	}

	err := validate.Struct(num)
	if err != nil {
		errs := err.(validator.ValidationErrors)
		for _, e := range errs {
			// can translate each error one at a time.
			fmt.Println(e.Translate(trans))
		}
	}
}
```
What did I expect to see?
```
num2 must be greater than num1
```
 What did I see instead
```
num2 must be greater than Num1
```
**tips**: `num1` is the tag value of struct field `Num1`

paulmil1:
+1, we definitely need this feature, its very important to be able to register 
RegisterParamNameFunc or something, similar to the
RegisterTagNameFunc

Tnr1112:
Hi, I wonder when the solution will be in production. Or if there is a way to implement it in my project temporarily. Thanks!

PR #601, opened 2020-04-22
enhance #597: get param field name with the tag name

cncal:
Enhances #597.

Change Details:

- Add `ParamField()` function in interface `FieldError` to get the **param field name** with the tag name taking precedence over the field actual name and this will also help with generating an error message. 

Usage:

For example, use `fe.ParamField()` instead of `fe.Param()` [here](https://github.com/go-playground/validator/blob/c68441b7f4748b48ad9a0c9a79d346019730e207/translations/en/en.go#L822).

@go-playground/admins


deankarn:
Hey @cncal I finally got some time to look into this deeper, nice work!

My only concerns are how this will affect the validations speed. It's a lot of extra work parsing for information that will only be used IF a validation fails. Is there a way that this could be changed to only extract that information if an error is encountered?

cncal:
@deankarn Sorry for replying you so late. I didn’t think of a good way to realize your idea and I suggest that tag this issue `help wanted` to find contributors who want to make some change based on this PR.

deankarn:
👍 going to close for now then

PR #614, opened 2020-05-23
Use ParamField() instead of Param() for default field validation error messages.

charleskorn:
Extends #601 to include updating default error messages for field validators to use the new function.

@go-playground/admins


deankarn:
Sorry for the confusion, I’m trying to wrap my head around how this is different from the existing Field() function available?

charleskorn:
Hi @deankarn, `ParamField()` effectively takes the `Param()` (which in the case of field validators is the field name to compare to), finds the corresponding field and applies the `TagNameFunc` to it. 

This means that error messages for these validators can then refer to the field with the same name as used in JSON, for example. 

So let's say we have this struct:

```golang
type Test struct {
    String1 string     `json:"string1"`
    String2 string     `json:"string2" validate:"eqfield=String1"`
}
```

and we've created a validator with a tag name function that extracts the JSON field name:

```golang
validate.RegisterTagNameFunc(func(fld reflect.StructField) string {
    name := strings.SplitN(fld.Tag.Get("json"), ",", 2)[0]

    if name == "-" {
        return ""
    }

    return name
})
```

If we then validate a `Test` value where `String2` is not the same as `String1`, without this change, the error message would be:

```
string2 must be equal to String1
```

With this change, the error message would use the JSON field name:

```
string2 must be equal to string1
```

I'm happy to add some tests to make this change clearer if that would be helpful.

FatalReadError:
This is fantastic, since the JSON that I'm getting from a web dashboard will be named the JSON fieldname (which is probably `phone_number` or something similar), not the exported Golang name `PhoneNumber`, and then I can match up the fieldnames in the front-end itself and show the appropriate error message directly below the error'ed field itself. It seems like the JSON name (such as `phonenum`) isn't accessible from the error codes at all.

deankarn:
closing until we can find an efficient way to implement this. original issue left open and marked with help wanted.

PR #616, opened 2020-05-28
Add isarr and notisarr validators

krhubert:
Enhances 

- [x ] Tests exist or have been written that cover this particular change.

Change Details:

Add two validators `inarr` and `notinarr` to check if a given value occurs in a slice.

Motivation:

Many times I need to make sure that one of the values is mutually exclusive in config/request. Example:

```
linters:
  enable:
    - import
  disable:
    - import
```
To make sure this can be validated I've created the above validators.

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -174,6 +174,8 @@ var (
 		"lowercase":            isLowercase,
 		"uppercase":            isUppercase,
 		"datetime":             isDatetime,
+		"inarr":                isInArr,
+		"notinarr":             isNotInArr,
 	}
 )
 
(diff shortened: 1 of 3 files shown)


baked_in.go
```diff
@@ -2096,3 +2098,39 @@ func isDatetime(fl FieldLevel) bool {
...
+		return false
+	}
+
+	for i := 0; i < arrv.Len(); i++ {
+		if arrv.Index(i).Interface() == field.Interface() {
+			return true
+		}
+	}
+
+	return false
```
Armatorix, line 2117: could be extracted (also below) and reused

deankarn:
@krhubert A couple of asks with this one, both relating to the validation name:

1. It would be great to come up with a more generic name for this validation and then extend it to maps as well.

2. Why about validation using a contact value rather than referencing a field? could we name it in such a way that denotes the validation param should be a field and not a value.

krhubert:
> It would be great to come up with a more generic name for this validation and then extend it to maps as well. could we name it in such a way that denotes the validation param should be a field and not a value.

That's a good idea.  About naming, I thought about `excludefiled`, `includefield`, but I don't like them... Maye you have some better names?

> Why about validation using a contact value rather than referencing a field? 

I don't understad what you mean here. Can you write more about `contact value`?

PR #630, opened 2020-07-03
Add checksum validation for Ethereum address

elias19r:
Fixes Or Enhances https://github.com/go-playground/validator/issues/631

Change Details:

- Add checksum validation for Ethereum address

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -531,17 +531,16 @@ func isEthereumAddress(fl FieldLevel) bool {
 	}
 
 	// Checksum validation. Reference: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md
-	address = address[2:]
+	address = address[2:] // Skip "0x" prefix.
 	h := sha3.NewLegacyKeccak256()
 	h.Write([]byte(strings.ToLower(address)))
 	hash := hex.EncodeToString(h.Sum(nil))
 
 	for i := 0; i < len(address); i++ {
-		if address[i] <= '9' {
+		if address[i] <= '9' { // Skip 0-9 digits: they don't have upper/lower-case.
 			continue
 		}
-
-		if hash[i] > '7' && address[i] >= 'a' || hash[i] <= '7' && address[i] < 'a' {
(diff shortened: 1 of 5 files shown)


baked_in.go
```diff
@@ -527,7 +530,28 @@ func isEthereumAddress(fl FieldLevel) bool {
...
+	// Checksum validation. Reference: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md
+	digitValue := map[byte]int{
+		'0': 0, '1': 1, '2': 2, '3': 3, '4': 4,
+		'5': 5, '6': 6, '7': 7, '8': 8, '9': 9,
+		'a': 10, 'b': 11, 'c': 12,
+		'd': 13, 'e': 14, 'f': 15,
+	}
+	address = address[2:]
+	h := sha3.NewLegacyKeccak256()
+	h.Write([]byte(strings.ToLower(address)))
```
elias19r, line 542: `hash.Hash`'s `io.Writer` implementation says it never returns an error. https://golang.org/pkg/hash/#Hash

baked_in.go
```diff
@@ -527,7 +530,28 @@ func isEthereumAddress(fl FieldLevel) bool {
...
+	hash := hex.EncodeToString(h.Sum(nil))
+
+	for i := 0; i < len(address); i++ {
```
elias19r, line 545: Here `len(address)` definitely has value 40, but I preferred to write `len(address)` instead of `40`. Thus we keep the information that "Ethereum addresses have 40 hex digits" declared only in one place, in the `regexes.go` (`ethAddress{Upper,Lower,}RegexString`).

Let me know if otherwise

baked_in.go
deankarn:
```suggestion
        // hash.Hash's io.Writer implementation says it never returns an error. https://golang.org/pkg/hash/#Hash
	_ , _ = h.Write([]byte(strings.ToLower(address)))
```

👆 this way linters don't complain and everyone knows the error checking wasn't just missed :)

deankarn:
@elias19r made one suggested change, other than that looks great and can merge next time I merge a group of PR's

deankarn:
@elias19r I'll do it during the merge no worries

Issue #626, opened 2020-06-12
validator treat json.Number as string


Tevic:
### Package version eg. v8, v9: 
v10

### Issue, Question or Enhancement:
validator treat json.Number as string

### Code sample, to showcase or reproduce:

```go
package main

import (
	"encoding/json"
	"fmt"

	"github.com/go-playground/validator/v10"
)

type Test struct {
	Val json.Number `validate:"min=20"`
}

func main() {
	validate := validator.New()
	test := Test{
		Val: "100",
	}
	fmt.Println(validate.Struct(test))
}


Output:
Key: 'Test.Val' Error:Field validation for 'Val' failed on the 'min' tag
Expect:
No Error
```

I know json.Number is string,but if we define a type as json.Number,we need this value act as a number value, so is there any way to let validator behavior act as number too?

elias19r:
Hi @Tevic

Yes, I think there is a way to let the validator act as a numeric value for `encoding/json.Number`

I've implemented it in PR https://github.com/go-playground/validator/pull/634

I'm not very knowledgeable about this codebase yet, but as [`encoding/json.Number`'s underlying type is `string`](https://golang.org/pkg/encoding/json/#Number), I think that creating a special case for `string` validations would make sense

Let's wait for maintainers to review it

deankarn:
Hey @Tevic @elias19r 

As much as I'd like to merge that PR `json.Number` is in fact a string and not a number as so I'm hesitant to add this special case because:
1. It may be a breaking change for those relying on it validating as a string already.
2. Treating this type as a special case would open up the door for every other type of special case and it's a door I don't want to open.

Having said that, this sort of thing has been thought of and there are 3 options for dealing with this sort of thing from least to most recommended:
1. Create a struct validation to handle this specific case if it's a one-off.
2. Create a custom validation that converts the value to an int for the validation.
3. Register a Custom Type Func. this is specifically designed to do what you want by converting the type to a different one prior to doing any validation. Here is an example of your use case: https://play.golang.org/p/MS_Dv4Nr3CE

Please let me know if this helps :)

PR #634, opened 2020-07-07
Add support for encoding/json.Number type (string)

elias19r:
Fixes Or Enhances https://github.com/go-playground/validator/issues/626

Change Details:

- Add support for `encoding/json.Number` type (`string`)
- Update related docs

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -175,6 +175,9 @@ var (
 		"uppercase":            isUppercase,
 		"datetime":             isDatetime,
 	}
+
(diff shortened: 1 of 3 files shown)


elias19r:
Given https://github.com/go-playground/validator/issues/626#issuecomment-654617342, it is preferred not to add a special case for `encoding/json.Number` inside the baked-in `string` validations. A user could register a custom function to deal with this special case

Issue #640, opened 2020-07-25
Add support for validating time zone name


math-nao:
### Package version eg. v8, v9: 
v10

### Issue, Question or Enhancement:
Enhancement
Add support for validating time zone name

This validates that a string value is a valid time zone based on the time zone database present on the system.
Although empty value and Local value are allowed by time.LoadLocation golang function, they are not allowed by this validator.
More information on https://golang.org/pkg/time/#LoadLocation

PR #641, opened 2020-07-25
Add support for time zone validation

math-nao:
Enhances #640 .

Change Details:

- Add new validator tag called "timeZone" to validate that a value is a valid time zone name
- Add test for this new validator
- Update documentation for this new validator

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -174,7 +174,7 @@ var (
 		"lowercase":            isLowercase,
 		"uppercase":            isUppercase,
 		"datetime":             isDatetime,
-		"timeZone":             isTimeZone,
+		"timezone":             isTimeZone,
 	}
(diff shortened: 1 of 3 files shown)


math-nao:
According this [topic](https://english.stackexchange.com/questions/3934/time-zone-vs-timezone) we can write both, `time zone` or `timezone` (mostly in US).

Moreover, according golang documentation, preferably spelling is two words `time zone`, so using camelCase it becomes `timeZone`. Documentation links: [FixedZone](https://golang.org/pkg/time/#FixedZone) [Time.Zone](https://golang.org/pkg/time/#Time.Zone).

If you still want to use single word spelling, confirm it to me, I will do the appropriate change.

PR #645, opened 2020-07-29
fix grammar/spelling mistakes in errors.go

NikolayOskin:
Fixes Or Enhances # .

Change Details:

- Fixing grammatical/spelling mistakes in function/field documentation in ```errors.go```

@go-playground/admins


andig:
Should be something like 

    if no registered translation can be found, it returns the original value

deankarn:
Updated the error message slightly different than this PR.

Issue #654, opened 2020-08-27
Panic Bad field type *string, when using required_without and omitempty at the same time


nigelis:
### Package version eg. v8, v9: 

v10.3.0

### Issue, Question, or Enhancement:

When validating a string pointer with `required_without` and `omitempty` at the same time, calling `Struct()` leads to panic.

Using `omitempty` merely is ok. Validating a string instead of a string pointer is ok.

### Code sample, to showcase or reproduce:

```go
package main

import (
	"log"

	"github.com/go-playground/validator/v10"
)

type Sample struct {
	Name string `validate:"omitempty,max=1024"`
	ID   string `validate:"required_without=Name,omitempty,max=64"`
}

type SampleWithPointer struct {
	Name *string `validate:"omitempty,max=1024"`
	ID   *string `validate:"required_without=Name,omitempty,max=64"`
}

func main() {
	name := "name"

	s := &Sample{Name: name}
	swp := &SampleWithPointer{Name: &name}

	validate := validator.New()
	if err := validate.Struct(s); err != nil {
		log.Printf("Validate sample with error, %v", err)
	}
	if err := validate.Struct(swp); err != nil {
		log.Printf("Validate SampeWithPointer with error, %v", err)
	}
}
```

nigelis:
Attach the stack trace.

```
panic: Bad field type *string

goroutine 1 [running]:
github.com/go-playground/validator/v10.isLte(0x11cdf20, 0xc0001245a0, 0xc00029fb00)
        /Users/nigel/go/pkg/mod/github.com/go-playground/validator/v10@v10.3.0/baked_in.go:1795 +0x63d
github.com/go-playground/validator/v10.hasMaxOf(0x11cdf20, 0xc0001245a0, 0xc0002afc01)
        /Users/nigel/go/pkg/mod/github.com/go-playground/validator/v10@v10.3.0/baked_in.go:1844 +0x35
github.com/go-playground/validator/v10.wrapFunc.func1(0x11ccee0, 0xc0000160a0, 0x11cdf20, 0xc0001245a0, 0x115a701)
        /Users/nigel/go/pkg/mod/github.com/go-playground/validator/v10@v10.3.0/baked_in.go:36 +0x39
[...]

PR #657, opened 2020-09-01
Validate tag omitempty, only base on field value

nigelis:
Fixes validation of tag omitempty, base on the field value #654.

Change Details:

- Validate the tag omitempty, only base on the field value.

@go-playground/admins

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -249,7 +249,7 @@ OUTER:
 			v.cf = cf
 			v.ct = ct
 
-			if !v.fldIsPointer && !hasValue(v) {
+			if !hasValue(v) {
 				return
 			}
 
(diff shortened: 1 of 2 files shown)


nigelis:
Hi @deankarn ,

Let me explain more about the change.
There are two code blocks in the function `traverseField` that validates the tag `omitempty`.

The first code block deals with the situation that `omitempty` is the first tag.
When the field value is a nil pointer, the `kind` would be `reflect.Ptr`, and no more validation would be continued.
https://github.com/go-playground/validator/blob/ea924ce89a4774b8017143b34b946db46add9df1/validator.go#L100-L111

The second code block deals with the situation that `omitempty` is not the first tag.
When the field value is a pointer, no matter its value, the validation is continued, which is not expected.
Because the `hasValue` can return the correct value, I simply remove the check about `fldIsPointer`.
https://github.com/go-playground/validator/blob/ea924ce89a4774b8017143b34b946db46add9df1/validator.go#L237-L257

All previous test cases are passed.
The added test case leads to panic in the v10.3.0, but are passed in the PR.

frederikhors:
This is still a problem. Is this released in v10? https://play.golang.org/p/TOBSAdFl-FC

PR #664, opened 2020-09-15
Closes #660 added excluded with and without

### LINKED ISSUE #660: Add support for denied_with and denied_without

v10

Add support for:
denied_with
denied_with_all
denied_without
denied_without_all

erma07:
Fixes #660 

Change Details:

- added baked_in.go excluded_with function (if there is x then there must not be a y)
- added baked_in.go excluded_with_all function
- added baked_in.go excluded_without function ( is that if x is not there then y must not be there )
- added baked_in.go excluded_without_all function
- added test cases

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -68,6 +68,10 @@ var (
 		"required_with_all":    requiredWithAll,
 		"required_without":     requiredWithout,
 		"required_without_all": requiredWithoutAll,
+		"excluded_with":        excludedWith,
+		"excluded_with_all":    excludedWithAll,
+		"excluded_without":     excludedWithout,
+		"excluded_without_all": excludedWithoutAll,
 		"isdefault":            isDefault,
 		"len":                  hasLengthOf,
 		"min":                  hasMinOf,
@@ -1383,6 +1387,18 @@ func requireCheckFieldKind(fl FieldLevel, param string, defaultNotFoundValue boo
 	}
 }
 
+// ExcludedWith is the validation function
+// The field under validation must not be present or is empty if any of the other specified fields are present.
+func excludedWith(fl FieldLevel) bool {
+	params := parseOneOfParam2(fl.Param())
+	for _, param := range params {
+		if !requireCheckFieldKind(fl, param, true) {
+			return !hasValue(fl)
+		}
+	}
+	return true
+}
+
 // RequiredWith is the validation function
 // The field under validation must be present and not empty only if any of the other specified fields are present.
 func requiredWith(fl FieldLevel) bool {
@@ -1393,6 +1409,18 @@ func requiredWith(fl FieldLevel) bool {
 		}
 	}
 	return true
+}
+
+// ExcludedWithAll is the validation function
+// The field under validation must not be present or is empty if all of the other specified fields are present.
+func excludedWithAll(fl FieldLevel) bool {
+	params := parseOneOfParam2(fl.Param())
+	for _, param := range params {
+		if requireCheckFieldKind(fl, param, true) {
+			return true
+		}
+	}
+	return !hasValue(fl)
 }
 
 // RequiredWithAll is the validation function
@@ -1407,6 +1435,15 @@ func requiredWithAll(fl FieldLevel) bool {
 	return hasValue(fl)
 }
 
+// ExcludedWithout is the validation function
+// The field under validation must not be present or is empty when any of the other specified fields are not present.
+func excludedWithout(fl FieldLevel) bool {
+	if requireCheckFieldKind(fl, strings.TrimSpace(fl.Param()), true) {
+		return !hasValue(fl)
+	}
+	return true
+}
+
 // RequiredWithout is the validation function
 // The field under validation must be present and not empty only when any of the other specified fields are not present.
 func requiredWithout(fl FieldLevel) bool {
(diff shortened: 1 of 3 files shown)


andig:
@erma07 see #665 for my attempt. Might be wrong but I guess you could steal the tests from there? Happy to close mine...

andig:
Ping @deankarn any chance this PR could get merged?

andig:
Is there any chance to consider this PR for merging?

andig:
Much appreciated, thank you!

baked_in.go
```diff
@@ -1416,6 +1453,18 @@ func requiredWithout(fl FieldLevel) bool {
 	return true
 }
 
+// RequiredWithoutAll is the validation function
```
jfrancisco0, line 1456: This should be `ExcludedWithoutAll`.

baked_in.go
```diff
@@ -1383,6 +1387,18 @@ func requireCheckFieldKind(fl FieldLevel, param string, defaultNotFoundValue boo
 	}
 }
 
+// ExcludedWith is the validation function
+// The field under validation must not be present or is empty if any of the other specified fields are present.
+func excludedWith(fl FieldLevel) bool {
```
jfrancisco0, line 1392: I'm trying to use `excluded_with` and not getting the expected behaviour. I only passes the validation if I set all the fields with the `excluded_with` validation, which is kind of the opposite of what I expected. Example:

```
type Example struct {
	AAA  *aaaStruct  `validate:"excluded_with=Bbb,omitempty,dive" json:"aaa,omitempty"`
	BBB  *bbbStruct  `validate:"excluded_with=AAA,omitempty,dive" json:"bbb,omitempty"`
}
```

If I set `AAA`, I get
```
"Key: 'Example.BBB' Error:Field validation for 'BBB' failed on the 'excluded_with' tag"
```

baked_in.go
andig: You could open a PR ;)

baked_in.go
andig: Do you have a play we could look at? Can you simplify it further (dive, omitempty) to expose the issue?

/cc @yaliv maybe you can spot the mistake?

baked_in.go
jfrancisco0: Sure, I'll do it tomorrow :)

baked_in.go
jfrancisco0: Ok, I'll try to confirm it and make sure I'm not messing up somewhere. I'll get back to you tomorrow!

yaliv: `excluded_with` validator works fine with basic/primitive fields (string, int).

I never set any validations for the sub-structs, because in my case I use `json.Unmarshal()` before validate.
Using `json.Unmarshal()` will always populate the sub-structs even if they do not present in the source JSON bytes.

This leads me to ask a fundamental question: "What is an empty struct value?"
- Is it `nil`?
- Is it `struct{}` (with zero fields)?

Unlike `map[string]interface{}`, which can hold "dynamic" fields, struct has a static set of fields.
If I define a struct with some fields in it (e.g. `MyStruct`), and then I create a struct value from it with its default (e.g. `var test MyStruct`), I still have access to those fields.

@jfrancisco0 if the built-in validators do not fit with your case, you can try creating some struct-level validations ([see example](https://github.com/go-playground/validator/blob/master/_examples/struct-level/main.go)). Plus, using this technique can minimize overhead, because you can validate multiple fields at once.

andig: @yaliv in this case the struct pointers are nil and hence definitely empty. The source code/docs confirm this.

jfrancisco0: I can confirm though that a field of type `bool` does not fail the validation, giving some strength to the hypothesis of structs behaving differently.

baked_in.go
jfrancisco0: https://github.com/go-playground/validator/pull/673

baked_in.go
yaliv: @andig if I write just `var test *MyStruct`, then `test` would be `nil`, but if I write `var test MyStruct`, then `&test` would not be `nil`.
So struct pointers have possible `nil` value.

Yes, I can see that fields with struct type and struct pointer type have a complete opposite behavior against primitive types.
Using the `excluded_with` validator on 2 fields, each mentioning each other:
- string type: both have values --> 2 validation errors; one has value --> validation passed :smiley:
- struct type: they always have values --> validation passed :frowning_face:
- struct pointer type: both have values --> validation passed; one has value --> 1 validation error; both empty --> 2 validation errors :frowning_face:

andig: so to summarise: there is an issue with struct pointers and excluded* validations. Sounds like it would be good to open a new issue with minimal example to reproduce. Might also be good to check required* behaviour for struct pointers too, as the pr code was a close copy.

andig: @yaliv did you by chance open follow-up issue?

yaliv: @andig sorry, I haven't done it yet.
I see this validator is quite complex, so I'm not sure if the problem is fixed it will not break other parts (e.g. the primitive types) or it will not cost performance.
But yes, it's still a good issue to be posted. Before it is fixed, maybe someone has a running workaround :wink:

Issue #707, opened 2020-12-24
can not use FieldError in ValidationErrors


xqbumu:
### Package version eg. v9, v10: 

v10

### Issue, Question or Enhancement:

can not use self defined FieldError with ValidationErrors.Translate.

### Code sample, to showcase or reproduce:

```go
package validation_test

import (
	"testing"

	"github.com/go-playground/locales/en"
	ut "github.com/go-playground/universal-translator"
	"github.com/go-playground/validator/v10"
	en_translations "github.com/go-playground/validator/v10/translations/en"
)

type FieldError struct {
	validator.FieldError
	ns string
}

func (fe FieldError) Namespace() string {
	return fe.ns
}

func TestTranslate(t *testing.T) {
	en := en.New()
	uni := ut.New(en, en)

	validate := validator.New()
	enTrans, _ := uni.GetTranslator("en")
	en_translations.RegisterDefaultTranslations(validate, enTrans)

	errs := validate.Var("", "min=1,max=10")
	if errs, ok := errs.(validator.ValidationErrors); ok {
		reErrs := validator.ValidationErrors{}
		for _, err := range errs {
			reErrs = append(reErrs, FieldError{err, "sfd"})
		}
		t.Log(reErrs.Translate(enTrans))
	}
}
```

### Error report

```
panic: interface conversion: validator.FieldError is validation_test.FieldError, not *validator.fieldError
```

PR #709, opened 2020-12-24
Fix FieldError in Translate

xqbumu:
issue introduced in #707

## Fixes Or Enhances

@go-playground/admins

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -44,12 +44,9 @@ func (ve ValidationErrors) Error() string {
 
 	buff := bytes.NewBufferString("")
 
-	var fe *fieldError
-
 	for i := 0; i < len(ve); i++ {
 
-		fe = ve[i].(*fieldError)
-		buff.WriteString(fe.Error())
+		buff.WriteString(ve[i].Error())
 		buff.WriteString("\n")
 	}
 
--- a/errors.go
+++ b/errors.go
@@ -61,10 +61,10 @@ func (ve ValidationErrors) Translate(ut ut.Translator) ValidationErrorsTranslati
 
 	trans := make(ValidationErrorsTranslations)
 
-	var fe *fieldError
+	var fe FieldError
 
 	for i := 0; i < len(ve); i++ {
-		fe = ve[i].(*fieldError)
+		fe = ve[i].(FieldError)
 
 		// // in case an Anonymous struct was used, ensure that the key
 		// // would be 'Username' instead of ".Username"
@@ -73,7 +73,7 @@ func (ve ValidationErrors) Translate(ut ut.Translator) ValidationErrorsTranslati
 		// 	continue
 		// }
 
-		trans[fe.ns] = fe.Translate(ut)
+		trans[fe.Namespace()] = fe.Translate(ut)
 	}
 
 	return trans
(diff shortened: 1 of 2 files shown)


deankarn (dismissed):
LGTM Thank You!

powerman:
Can you please make similar change in ValidationErrors.Error? Forced type conversion to concrete unexported type prevents using mocks when testing for returned errors.

xqbumu:
> 
> 
> Can you please make similar change in ValidationErrors.Error? Forced type conversion to concrete unexported type prevents using mocks when testing for returned errors.

Maybe we can delete the fe in ValidationErrors.Error ? @powerman

powerman:
> Maybe we can delete the fe in ValidationErrors.Error ? @powerman

Yep, `ve[i].Error()` should works just fine.

powerman:
@xqbumu Thanks!
@deankarn Can you please merge this? (BTW, you probably should add 1.16 to testing matrix.)

xqbumu (commented):
LGTM

deefdragon:
Not sure if this is still close enough to be merged/merged with minor edits, but would it be possible for this to also cover [this type assertion](https://github.com/go-playground/validator/blob/9e2ea4038020b5c7e3802a21cfa4e3afcfdcd276/struct_level.go#L163-L175)?

If not, I would be happy to make a PR that (tries to) address that individually/all these issues at once.

errors.go
```diff
@@ -61,10 +58,10 @@ func (ve ValidationErrors) Translate(ut ut.Translator) ValidationErrorsTranslati
 
 	trans := make(ValidationErrorsTranslations)
 
-	var fe *fieldError
+	var fe FieldError
 
 	for i := 0; i < len(ve); i++ {
-		fe = ve[i].(*fieldError)
+		fe = ve[i].(FieldError)
```
joselitofilho, line 64: It is not necessary to make this cast here. The `ve` is already a slice of `FieldError`.
```suggestion
		fe = ve[i]
```

joselitofilho:
I'm getting the same problem. Please, merge this.

errors.go
xqbumu: Thanks for your review. I have removed it with test.

msaboor35:
`ReportValidationErrors` currently depends on casting `FieldError` to the unexported `*fieldError` type to prepend current namespaces to the relative ones. This makes it impossible to use this function outside the library with custom `FieldError` implementations.

Can we revisit this in the current PR and consider a fix, so it works without the unexported type `fieldError`? happy to contribute a fix, so we can get this unblocked and merged.

deankarn:
This PR is pretty old @msaboor35 

is this still an issue? Also exposing the in ere type like this would likely not play nice with errors needing to comply with FieldError interface.

if it is an issue still the solution would need to ensure errors comply with the interface and not sure exposing the inner error is the right way vs an interface approach.

msaboor35:
This was fixed for ValidationErrors' `Error` method in #981 but the problem still exists in ValidationErrors' `Translate` method and `ReportValidationErrors` method.

I also don't think we should be exposing internal `fieldError` type. We must rely on the exported `FieldError` interface. This current PR also doesn't expose the `fieldError` for `Translate` fix and the fix I have in mind for `ReportValidationErrors` also doesn't expose it.

Since it is pretty old PR, I can open a new PR, if desired.

PR #710, opened 2020-12-26
README.md - version removed from installation and import

Serj1c:
from installation and import parts of the readme file "/v10" was removed (go get did not work with it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ Installation
 
 Use go get.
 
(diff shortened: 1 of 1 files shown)


deankarn (approved):
TY!

ByungJun25:
For me, ` go get github.com/go-playground/validator/v10` is correct. If I use `go get github.com/go-playground/validator`, it takes `v9.31.0`...

Which command is right?

PR #730, opened 2021-02-20
feat: add bcp 47 language tag validator

joshuaslate:
## Fixes Or Enhances
This adds a new validator: `language_tag`, which uses x/language to parse a string to validate that it's a BCP 47 tag.

@go-playground/admins


baked_in.go
```diff
@@ -188,6 +189,7 @@ var (
 		"iso3166_1_alpha2":        isIso3166Alpha2,
 		"iso3166_1_alpha3":        isIso3166Alpha3,
 		"iso3166_1_alpha_numeric": isIso3166AlphaNumeric,
+		"language_tag":            isLanguageTag,
```
deankarn, line 192: @joshuaslate what do you think about changing this to `bcp47_language_tag` to be specific and future proof for adding others?

joshuaslate: Thank you for the review, @deankarn. Good call-- I've updated the PR 😄

PR #733, opened 2021-02-25
Fix accessing fields in composed structs

Adimus11:
## Fixes Or Enhances

This change fixes the issue with validating fields in composed structures when the validator uses `.` before the composed struct in the parent one.

@go-playground/admins

### DIFF
--- a/validator_instance.go
+++ b/validator_instance.go
@@ -409,8 +409,8 @@ func (v *Validate) StructPartialCtx(ctx context.Context, s interface{}, fields .
 		if len(flds) > 0 {
 
 			vd.misc = append(vd.misc[0:0], name...)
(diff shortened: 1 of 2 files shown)


validator_instance.go
```diff
@@ -409,7 +409,10 @@ func (v *Validate) StructPartialCtx(ctx context.Context, s interface{}, fields .
 		if len(flds) > 0 {
 
 			vd.misc = append(vd.misc[0:0], name...)
-			vd.misc = append(vd.misc, '.')
+			// Don't append empty name for composed structs
+			if string(vd.misc) != "" {
```
deankarn, line 413: Can we check the length of vd.misc here instead of casting to a string? It would be one less allocation :)

PR #746, opened 2021-03-31
Add support for excluded_with on pointer fields

### LINKED ISSUE #745: exclude_with and exclude_without don't work on pointer fields
v10

Using the `exclude_with` or `exclude_without` tags don't work as expected on pointer fields (e.g. `*string`). 

See the example below, where the validation fails even though the "excluded" fields are never included. Change the types from `*string` to `string` and now the validation succeeds.

```go
type Test1 struct {
	Field1 string
	Field2 *string `validate:"excluded_with=Field1"`
}

type Test2 struct {
	Field1 string
	Field2 *string `validate:"excluded_without=Field1"`
}

func main() {
	v := validator.New()
	fmt.Println(v.Struct(&Test1{Field1: "ok"})) // Fails, even though Field2 is null
	fmt.Println(v.Struct(&Test2{Field1: "ok"})) // Fails, even though Field2 is null
}
```

https://play.golang.org/p/QTjxLJOSF8W

Jake-Convictional:
## Fixes Or Enhances
Closes #745

Using a tag like `excluded_with` will now work as expected on pointer fields (e.g. `*string`). There are two commits here. The first commit adds in tests (which fail) to demonstrate the behavior that was missing. The second commit improves the code so that the tests pass.

I checked to see if this would also address [this comment](https://github.com/go-playground/validator/pull/664#discussion_r500934739) about struct pointers, but it does not.

@go-playground/admins

### DIFF
--- a/validator_test.go
+++ b/validator_test.go
@@ -10202,6 +10202,28 @@ func TestExcludedWith(t *testing.T) {
 		name := fmt.Sprintf("Field%d", i)
 		AssertError(t, errs, name, name, name, name, "excluded_with")
 	}
+
+	test3 := struct {
+		Inner  *Inner
+		Inner2 *Inner
+		Field  string            `validate:"omitempty" json:"field"`
+		FieldE string            `validate:"omitempty" json:"field_e"`
+		Field1 string            `validate:"excluded_with=FieldE" json:"field_1"`
+		Field2 *string           `validate:"excluded_with=FieldE" json:"field_2"`
+		Field3 map[string]string `validate:"excluded_with=FieldE" json:"field_3"`
+		Field4 interface{}       `validate:"excluded_with=FieldE" json:"field_4"`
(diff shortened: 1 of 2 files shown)


validator_test.go
```diff
@@ -10352,9 +10417,10 @@ func TestExcludedWithoutAll(t *testing.T) {
 		Field2 *string           `validate:"excluded_without_all=Field FieldE" json:"field_2"`
 		Field3 map[string]string `validate:"excluded_without_all=Field FieldE" json:"field_3"`
 		Field4 interface{}       `validate:"excluded_without_all=Field FieldE" json:"field_4"`
-		Field5 string            `validate:"excluded_without_all=Inner.Field Inner.Field2" json:"field_5"`
+		Field5 string            `validate:"excluded_without_all=Inner.Field Inner2.Field" json:"field_5"`
```
Jake-Convictional, line 10420: This seemed like a typo

PR #752, opened 2021-04-15
Add ability to validate map data

zytell3301:
Add ability to validate map data based on map validation rules.
In some frameworks like kataras/iris, they return a map of form values. 
So it is hard to validate data that is not in struct form. So we can easily provide the data
itself and validation rules in form of map to have an easier data validation

@go-playground/admins

### DIFF
--- a/_examples/map-validation/main.go
+++ b/_examples/map-validation/main.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+	"fmt"
+	"github.com/go-playground/validator/v10"
+)
+
+var validate *validator.Validate
+
+func main() {
+	validate = validator.New()
+
+	validateMap()
+}
+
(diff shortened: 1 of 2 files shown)


deankarn:
Thanks for the PR @zytell3301 

My only request would be that it support nested data. This looks great for flat data structures but not sure how you could apply validations for nested data.

Thoughts?

zytell3301:
I think i've found a solution for it. I'll push the changes after I implemented it.

Thanks for your attention @deankarn and apologies for any mistakes in PR or commits.
I'm new to these stuff

zytell3301:
Hi. I just implemented validating nested maps. But cross-field validation and keys validation is missing for now. Please check the updates.

deankarn (approved):

PR #753, opened 2021-04-16
Creating custom regex validation

LucasMendesl:
## Fixes Or Enhances

Creating custom `regex` validation.

@go-playground/admins

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"os"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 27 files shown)


wpliang:
this function is great, does this available now?

baked_in.go
```diff
@@ -767,6 +769,11 @@ func fieldExcludes(fl FieldLevel) bool {
 	return !strings.Contains(field.String(), currentField.String())
 }
 
+func regex(fl FieldLevel) bool {
+	regex := regexp.MustCompile(fl.Param())
```
rogchap, line 773: Running `MustCompile` on every validation is expensive (and an anti-pattern). Is there a way to call compile once and cache the regexp?

PR #759, opened 2021-04-25
Postcode validator for value and field

vuon9:
## Rules for postcode validation
- `postcode_iso3166_alpha2`: Use directly with a value of country code in ISO 3316 alpha 2
- `postcode_iso3166_alpha2_field`: Use directly with a field which should contain a value of country code in ISO 3316 alpha 2

## Notes
- Original discussion https://github.com/go-playground/validator/discussions/751
- It does not support ISO 3316 other types, and not for all countries.
- Added postcode regex for Vietnam (VN) country

@go-playground/admins


deankarn:
Thanks for the PR @vuon9 I'll try and get to it soon. Apologies it might take a while.

deankarn (approved):

PR #761, opened 2021-04-27
update: _examples/simple/main.go ambiguous naming

huoyinghui:
## Fixes

@go-playground/admins

I'm new to golang, and I think using err is better than errs, because  validate.Var return error.

and, i test in my goland.

### DIFF
--- a/_examples/simple/main.go
+++ b/_examples/simple/main.go
@@ -90,10 +90,10 @@ func validateVariable() {
 
 	myEmail := "joeybloggs.gmail.com"
 
-	errs := validate.Var(myEmail, "required,email")
+	err := validate.Var(myEmail, "required,email")
 
-	if errs != nil {
-		fmt.Println(errs) // output: Key: "" Error:Field validation for "" failed on the "email" tag
+	if err != nil {
+		fmt.Println(err) // output: Key: "" Error:Field validation for "" failed on the "email" tag
 		return
 	}
 


deankarn:
@huoyinghui I did this in order to keep the outer and inner errors differentiated to avoid confusion that their variable names were the same.

your changes do work as the inner `err` shadows the outer one, but I'd prefer to keep them with different names to avoid the confusion. I'm not apposed to changing the name though if it makes it clearer.

huoyinghui:
@deankarn ,  Are you referring to the code of this block? you are right! the var of err in this code,  error has be used twice

_examples/simple/main.go:56
```go
	// returns nil or ValidationErrors ( []FieldError )
	err := validate.Struct(user)
	if err != nil {

		if _, ok := err.(*validator.InvalidValidationError); ok {
			fmt.Println(err)
			return
		}

			fmt.Println(err.Namespace())
			fmt.Println(err.Field())
			fmt.Println(err.StructNamespace())
			fmt.Println(err.StructField())
			fmt.Println(err.Tag())
			fmt.Println(err.ActualTag())
			fmt.Println(err.Kind())
			fmt.Println(err.Type())
			fmt.Println(err.Value())
			fmt.Println(err.Param())
			fmt.Println()
		}

		// from here you can create your own error messages in whatever language you wish
		return
	}
```


in this code(_examples/simple/main.go:89), i think errs only local, not shadows the outer one. just use err is good
```go
func validateVariable() {

	myEmail := "joeybloggs.gmail.com"

	errs := validate.Var(myEmail, "required,email")

	if errs != nil {
		fmt.Println(errs) // output: Key: "" Error:Field validation for "" failed on the "email" tag
		return
	}

	// email ok, move on
}
```

Issue #711, opened 2020-12-30
How do you add a custom validator and error message?


Xeoncross:
### Package version eg. v9, v10: 

v10

It seems adding a custom validation rule is something that should be pretty common. I've been unable to find a complete example of adding a validation rule and the matching error message. Could someone explain how this is to be done?

https://github.com/go-playground/validator/blob/d6b17fd90bd4dd9d16a594c3035ceadc3de0193a/doc.go#L1273-L1282

```go
	type Test struct {
		TestField string `validate:"yourtag"`
	}
	
	validate := validator.New()
	validate.RegisterValidation("yourtag", validators.NotBlank)
	
	// now how do you set the error message?
```

It looks like you might have to add this [non-intuitive block of code to mutate the translation handling](https://github.com/go-playground/validator/blob/f6584a41c8acc5dfc0b62f7962811f5231c11530/_examples/translations/main.go#L105-L111)

```go
	validate.RegisterTranslation("yourtag", trans, func(ut ut.Translator) error {
		return ut.Add("yourtag", "{0} must have a value!", true) // see universal-translator for details
	}, func(ut ut.Translator, fe validator.FieldError) string {
		t, _ := ut.T("yourtag", fe.Field())
		return t
	})
```

Related

- https://github.com/go-playground/validator/issues/669

megamayoy:
Yes, this is how it's done. For a more clarifying example , refer to this blog post https://medium.com/@apzuk3/input-validation-in-golang-bc24cdec1835

Issue #669, opened 2020-09-28
Custom error message


1379:
The parameter of the RegisterValidation function is a validation function that returns a Boolean value. If we want to customize the error message, we have to register a translation function, which is too much trouble. 
Our verification function may return an error for many reasons, so the error message should also be different, but only based on the returned boolean value, we cannot determine what caused the verification failure, so the error message of the translation function is only One kind.
Why can't you register a check function that returns error directly?
Just like the following.

```
type Func=func(fl FieldLevel) error
func (v *Validate) RegisterValidation(tag string, fn Func, callValidationEvenIfNull ...bool) error {

}

```

1379:
Only based on the boolean value returned by the validation function, we cannot determine what caused the validation failure, so customizing the error message will be very troublesome

deankarn:
@1379 sounds like a great enhancement for the next version of the lib, but unfortunately right now is a breaking change.

1379:
@deankarn 
I simply made some changes to the "validator", which is currently supported and used in my own project.
My implementation is like this:
Provides two RegisterValidation functions, the first parameter type is still "type Func=func(fl FieldLevel) boolean",
The parameter type of the second function becomes "type Func=func(fl FieldLevel) error".
When verifying, you only need to determine which type of verification function is.

pwli0755:
Same need

kke:
The custom validation error messages from `sl.ReportError()` are a bit silly too:

```
// sl.ReportError(field interface{}, fieldName string, structFieldName string, tag string, param string)
sl.ReportError(hosts, "hosts", "this does nothing", "some error message here", "this does nothing")
```

==> 

```
 Key: 'ClusterConfig.Spec.hosts' Error:Field validation for 'hosts' failed on the 'some error message here' tag
```

If it was not from a tag, but a custom StructValidation , it should say something like:

```
 Key: 'ClusterConfig.Spec.hosts' Error:Field validation for 'hosts' failed: some error message here
```

yaofei517:
here is a way to fix it #837

pbikki:
same need. especially for `sl.ReportError()` on struct validation to return custom error messages to user

joostvdhoff:
I think it's possible to implement this without a breaking change, though it'll be ugly with some bits of code purely dedicated to backwards compatibility. We'd have to put a new registration mechanism with new names next to the current one, and turn the current one into a proxy:

```go
type FuncWithCustomError=func(fl FieldLevel) error
[...]

PR #787, opened 2021-06-23
support custom error message

qm012:
#### Now the default error message is displayed
- default 
`
Key: 'User.Email' Error:Field validation for 'Email' failed on the 'email' tag
`
- RegisterTranslation
`
validator.New().RegisterTranslation()
`
#### It's too complicated and painful

relevant issues:
#711
#669


qm012:
hi @DavidLarsKetch Can you please review this? thanks

deankarn:
Sorry for taking so long, it'll be a while before I can review properly.

qm012:
Ok. Thank you for your review.:cow2:

odannyc:
This would definitely be useful!

validator_test.go
```diff
@@ -11280,3 +11280,67 @@ func TestPostCodeByIso3166Alpha2Field_InvalidKind(t *testing.T) {
 	_ = New().Struct(test{"ABC", 123, false})
 	t.Errorf("Didn't panic as expected")
 }
+
+func TestCustomMessage(t *testing.T) {
+	type test struct {
+		Min      int    `validate:"min=1" message:"test int min"`
```
smithaitufe, line 11286: I think I prefer something like this

```
type Test struct {
Min int `validate:"min=1[test int min],max=10[test int max]"`
}
```
If you can implement this, then it will be beautiful.

ipfans:
Hello guys. Do we have any progress on this PR?

PR #795, opened 2021-07-16
Add test for not required e164 (phonenumber)

genofire:
## Fixes
or better create an Test to make it visible an fix is necessary


deankarn:
I'm not sure what this PR is attempting to show. Is `omitempty` not what you are looking for?

genofire:
Do you mean `json:"omitempty"` ? This just has impact of Marshalling ... It is still jnternal an empty string.

It just show, that the current regex solution also run on empty values. On this way it is an indirect `required` and a useless library for optional values.

deankarn:
@genofire  please read the documentation, omitempty is what your are missing if you want empty or a valid value

validator_test.go
```diff
@@ -5515,7 +5515,7 @@ func TestStructOnlyValidation(t *testing.T) {
 		LastName          string     `json:"lname"`
 		Age               uint8      `validate:"gte=0,lte=130"`
 		Number            string     `validate:"required,e164"`
-		AlternativeNumber string     `validate:"e164"`
+		AlternativeNumber string     `validate:"e164,omitempty"`
```
deankarn, line 5517: This is NOT how omit empty work with validator. PLEASE read the documentation provided.

`omitempty` comes before whatever you want that to be true for.

genofire: Where is it dritten, that `omitempty` needs to be before other validators?

```
Allows conditional validation, for example if a field is not set with a value (Determined by the "required" validator) then other validation such as min or max won't run, but if a value is set validation will run.
```

genofire:
Why is it closed?

PR #814, opened 2021-08-08
fix: Fixed panic when use ru lang for translation, added and fixed tests

anisov:
## Panic when using Russian for translations of some tags with values ​​other than 1

When we use:
```
ru := russian.New()
uni := ut.New(ru, ru)
trans, _ := uni.GetTranslator("ru")
```
We get panic when validating structures with tags min, len, etc. You didn't get an error in your tests, because used:
```
russian "github.com/go-playground/locales/en"
```
instead of  
```
russian "github.com/go-playground/locales/ru"
```

Panic arises because in the locales package for the Russian language the CardinalPluralRule function differs in its implementation of the function in English.

RU:
```
// CardinalPluralRule returns the cardinal PluralRule given 'num' and digits/precision of 'v' for 'ru'
func (ru *ru) CardinalPluralRule(num float64, v uint64) locales.PluralRule {

	n := math.Abs(num)
	i := int64(n)
	iMod10 := i % 10
	iMod100 := i % 100

	if v == 0 && iMod10 == 1 && iMod100 != 11 {
		return locales.PluralRuleOne
	} else if v == 0 && iMod10 >= 2 && iMod10 <= 4 && (iMod100 < 12 || iMod100 > 14) {
		return locales.PluralRuleFew
	} else if (v == 0 && iMod10 == 0) || (v == 0 && iMod10 >= 5 && iMod10 <= 9) || (v == 0 && iMod100 >= 11 && iMod100 <= 14) {
		return locales.PluralRuleMany
	}

	return locales.PluralRuleOther
}
```
EN:
```
// CardinalPluralRule returns the cardinal PluralRule given 'num' and digits/precision of 'v' for 'en'
func (en *en) CardinalPluralRule(num float64, v uint64) locales.PluralRule {

	n := math.Abs(num)
	i := int64(n)

	if i == 1 && v == 0 {
		return locales.PluralRuleOne
	}

	return locales.PluralRuleOther
}
```
This led to panic in the program when we used some tags with numbers other than 1. 
It should also be borne in mind that the suffix at the end of words in Russian changes not only depending on the number in front, but also on the structure of the sentence, therefore they are not the same for different tags and additional tests have been written for this. And if the structure of the message output changes in the future, then the suffixes at the end of words may also change, but at the moment they are correct.
It is also worth noting that in this context locales.PluralRuleOther will never drop out in Russian, but I left the implementation in the ru module, because the function can return the given value.


deankarn:
Thanks @anisov for fixing this.

PR #828, opened 2021-09-07
validator_instance.go line 300 confict of map

liuyongchen:
Concurrent map reads and writes occur randomly

## Fixes Or Enhances

Add read/write locks to the structure to solve this problem


deancn (commented):
LGTM

deankarn:
@liuyongchen I am not sure issue this PR is trying to solve?

The addition of this mutex would only guard the RegisterTranslation function suggesting you’re calling it multiple times in parallel or during your application runtime.

RegisterTranslation should only be called during setup/config of your application and so has no need to be thread safe.

please file an issue if there is a use case I’m missing but right now see no need for this PR

deankarn:
Translation should only be registered upon config, never at runtime.

PR #839, opened 2021-09-21
add the error msg tag

yaofei517:
## add a error msg tag

this tag will effect to the error defind after this tag,
struct filed and dive will be also effected


yaofei517:
review test

yaofei517 (commented):
review test

PR #847, opened 2021-10-14
added excluded_if/excluded_unless tags + tests

jh125486:
## Fixes Or Enhances

Adds `excluded_if` and `excluded_unless` tags with tests.


agbishop:
This would be useful, can someone please review this?

baked_in.go
```diff
@@ -1558,6 +1576,22 @@ func requiredUnless(fl FieldLevel) bool {
...
 }
 
+// excludedUnless is the validation function
+// The field under validation must not be present or is empty unless all the other specified fields are equal to the value following with the specified field.
+func excludedUnless(fl FieldLevel) bool {
+	params := parseOneOfParam2(fl.Param())
+	if len(params)%2 != 0 {
+		panic(fmt.Sprintf("Bad param number for excluded_unless %s", fl.FieldName()))
+	}
+	fmt.Println(params[0], params[1], requireCheckFieldValue(fl, params[0], params[1], false), !hasValue(fl))
```
deankarn, line 1586: Leftover debug?

deankarn (commented):
Looking great overall, does it work with nested fields though?

eg. If one wanted to require unless field in a parent or child structs field?

baked_in.go
jh125486: Yep, 100% a leftover.   Pushed with changes.
I'll add a test for nested fields today (hopefully).

jh125486: OK, tests for nested (inner) structs were added.

Golangci-lint is very angry though, but locally runs fine (I'm running 1.45.2 though). 
I also don't see a golangci lint yaml config in the root dir...

deankarn (approved):

PR #883, opened 2022-01-13
validate top level slice

### LINKED ISSUE #882: top level slice of structs

We have a use case where we want to validate slice of structs. We cannot wrap the slice in a struct, because we need to return json array (not json object with array field). Current validator allows only validation if the input (top) is a struct (I know that we can add `dive` tag, but it is not possible to add tag to a slice e.g.:

```go
type Output []SomeStruct `validate"required,dive"` // does not compile, tags can be added only on struct type
---
type Output []SomeStruct
type OutputStruct struct {
    Output `validate"required,dive"` // json will wrap slice in "output", "output" is omitted only if the type is struct
}
```

```go
	in := []struct {
		Value string `validate:"contains=@"`
	}{
		{Value: "test@1"}, {Value: "test@2"},
	}
	validate := New()
	errs := validate.Struct(in) // fails with InvalidValidationError
```
I understand that slice and map can be validated with `dive`, but as mentioned above it is imposible to add `dive` tag on the slice itself and we cannot wrap slice in a struct, because we need to return (json) array.

pete911:
## Fixes Or Enhances

Adds ability to validate top level slice of structs, fixes #882 .

### DIFF
--- a/validator_instance.go
+++ b/validator_instance.go
@@ -325,6 +325,30 @@ func (v *Validate) Struct(s interface{}) error {
 func (v *Validate) StructCtx(ctx context.Context, s interface{}) (err error) {
 
 	val := reflect.ValueOf(s)
+	if val.Kind() == reflect.Ptr && !val.IsNil() {
+		val = val.Elem()
+	}
+
+	// input is not a slice
+	if val.Kind() != reflect.Slice {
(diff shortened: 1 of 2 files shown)


deankarn:
see https://github.com/go-playground/validator/issues/882#issuecomment-1114281307

Issue #910, opened 2022-03-12
Result is wrong while there are multiple group of OR operators


XIELongDragon:
- [x] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [x] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version eg. v9, v10: 

v9 and v10

### Issue, Question or Enhancement:
If there are multiple group of OR operators in the tag, the validation result is wrong

### Code sample, to showcase or reproduce:

```go
type T struct {
   Value int `validate:"eq=1|gte=5,eq=1|lt=7"`
}

t := T{8}
```

The validation condition we set is `(Value==1 || Value >=5) && (Value==1 || Value <7)` , so if the `Value=8` the validate should faild, but it passes the test

deankarn:
PR #910 merged

PR #911, opened 2022-03-12
Result is wrong while there are multiple group of OR operators #910

XIELongDragon:
## Fixes Or Enhances
Fix bug #910

```go
type T struct {
   Value int `validate:"eq=1|gte=5,eq=1|lt=7"`
}

t := T{8}
```

should failed the test


deankarn (approved):
LGTM, thanks for the PR!

I'll try to merge the next time I do a release.

PR #912, opened 2022-03-15
fix: Remove underscore from RFC-1123 based regex

sdillen:
RFC-1123 is based on RFC-952, which doesn't allow underscores. RFC-1123
must be therefore implemented with the same constraint to disallow
underscores in host names.

## Fixes Or Enhances

- Regex that implement RFC-1123 don't allow underscores anymore


zemzale (changes requested):
Thanks for the PR @sdillen 

Could you add a failing test case for the hostname that would contain the `_` so we have that covered and we wouldn't have regression in the future? 

It should be somewhere in `validator_test.go` test case `TestHostnameRFC1123Validation`

jeffwidman (approved):

PR #924, opened 2022-03-25
Credit card validation

alessmar:
Implemented Luhn algoritm to validate credit card string

## Fixes Or Enhances

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -2438,26 +2438,27 @@ func isDnsRFC1035LabelFormat(fl FieldLevel) bool {
 	return dnsRegexRFC1035Label.MatchString(val)
 }
 
+// isCreditCard is the validation function for validating if the current field's value is a valid credit card number
 func isCreditCard(fl FieldLevel) bool {
 	val := fl.Field().String()
 	var creditCard bytes.Buffer
-	parts := strings.Split(val, " ")
-	for _, part := range parts {
-		if len(part) < 2 {
+	segments := strings.Split(val, " ")
+	for _, segment := range segments {
+		if len(segment) < 3 {
 			return false
 		}
-		creditCard.WriteString(part)
+		creditCard.WriteString(segment)
 	}
 
 	ccDigits := strings.Split(creditCard.String(), "")
 	size := len(ccDigits)
-	if size < 14 || size > 16 {
+	if size < 12 || size > 19 {
 		return false
 	}
 
 	sum := 0
-	for i := 0; i < size; i++ {
-		value, err := strconv.Atoi(ccDigits[i])
+	for i, digit := range ccDigits {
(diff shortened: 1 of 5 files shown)


baked_in.go
```diff
@@ -2436,3 +2437,40 @@ func isDnsRFC1035LabelFormat(fl FieldLevel) bool {
...
+	for _, part := range parts {
+		if len(part) < 2 {
+			return false
+		}
+		creditCard.WriteString(part)
+	}
+
+	ccDigits := strings.Split(creditCard.String(), "")
+	size := len(ccDigits)
+	if size < 14 || size > 16 {
```
marrow16, line 2454: Globally, card numbers (PAN) can be 12 - 19 digits
(see https://en.wikipedia.org/wiki/Payment_card_number)

alessmar: @marrow16 thank you for pointing out the min and max size of a PAN. I've updated the code according to the documentation that you provided.

grandeto: @marrow16 @alessmar 
Hi folks,
Actually PAN length is in range 10-19

`IIN length has been extended to 8-digits in fifth edition of ISO/IEC 7812 published in 2017[[2]](https://en.wikipedia.org/wiki/Payment_card_number#cite_note-2) and PAN will continue to remain variable length, ranging from 10 to 19 digits.`

baked_in.go
```diff
@@ -2436,3 +2437,41 @@ func isDnsRFC1035LabelFormat(fl FieldLevel) bool {
...
+	for _, segment := range segments {
+		if len(segment) < 3 {
+			return false
+		}
+		creditCard.WriteString(segment)
+	}
+
+	ccDigits := strings.Split(creditCard.String(), "")
+	size := len(ccDigits)
+	if size < 12 || size > 19 {
```
grandeto, line 2455: Hi,
I wrote this comment below the previous comments' thread. 
However just in case putting it also here

The correct PAN validation nowadays should be `size < 10 || size > 19`

Here is the official statement:

https://www.iso.org/news/2016/11/Ref2146.html

Within the current version of ISO/IEC 7812-1, an IIN is defined as a fixed-length numeric of six digits. The standard also defines the Primary Account Number (PAN), a number which is used to identify an individual account holder. The PAN is of variable length, ranging from 8 to 19 digits.

Due to the increasing number of card issuers, there is expected to be a shortage in the available supply of IINs. Therefore, ISO/IEC 7812-1 is being revised to expand the IIN to an eight-digit numeric value from the current six digits. The PAN will continue to remain a variable length, `ranging from 10 to 19 digits.`

grandeto (changes requested):

deankarn (approved):
👍

Issue #853, opened 2021-11-11
How to validate a struct without tags?


chenyanchen:
- [x] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [x] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version v10: 

`import "github.com/go-playground/validator/v10"`

### Issue, Question or Enhancement:

I made some gRPC services with protobuf, the generated files has no struct tags. How to validate a struct without struct tags?

Maybe [gogo/protobuf](https://github.com/gogo/protobuf) can add some struct tags into the generation files, but this is no matter to use a new package for the complex method.

I find the same problem #722 not same pain spot and closed.

There is some simply method to solove this problem?

### Code sample, to showcase or reproduce:

`.proto` file
``` proto
message SearchRequest {
  string RequestID = 1;
  uint64 UserID = 2;
  string Email = 3;
  repeated string Keywords = 4;
}
```
Generated structure:
```go
type SearchRequest struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	RequestID string   `protobuf:"bytes,1,opt,name=RequestID,proto3" json:"RequestID,omitempty"`
	UserID    uint64   `protobuf:"varint,2,opt,name=UserID,proto3" json:"UserID,omitempty"`
	Email     string   `protobuf:"bytes,3,opt,name=Email,proto3" json:"Email,omitempty"`
	Keywords  []string `protobuf:"bytes,4,rep,name=Keywords,proto3" json:"Keywords,omitempty"`
}
```
I want validate this structure equal this:
```go
type SearchRequest struct {
	RequestID string `validate:"required"`
	UserID    uint64
	Email     string   `validate:"omitempty,email"`
	Keywords  []string `validate:"dive,required"`
}
```
In the proto generated files, there is no tags, so I expect a method like `RegisterStructMap`, the `map[string]string` is `map[FieldName]tag`.
```go
func main() {
	v := validator.New()
	v.RegisterStructMap(
		SearchRequest{}, map[string]string{
			"RequestID": "required",
			"Email":     "omitempty,email",
			"Keywords":  "dive,required",
		},
	)
	req := &SearchRequest{
		Keywords: []string{""},
	}
	if err := v.Struct(req); err != nil {
		fmt.Println(err)
	}
}
// Output:
[...]

PR #934, opened 2022-04-20
Feat: support validate struct without struct tag

leoliang1997:
## Fixes Or Enhances

validate struct without struct tag, use `map[string]string` to define rules, ref: #853 


validator_instance.go
```diff
@@ -271,6 +272,33 @@ func (v *Validate) RegisterStructValidationCtx(fn StructLevelFuncCtx, types ...i
 	}
 }
 
+// RegisterStructValidationMapRules registers validate map rules
```
deankarn, line 275: Can you add some addition information regarding that rules override any tags at the struct level?

deankarn (approved):
Very cool just one ask.

validator_instance.go
deankarn: Ah I can add it.

PR #950, opened 2022-05-24
Add ObjectID validator

smgt:
## Fixes Or Enhances

Add a validator for [ObjectIDs](https://github.com/mongodb/specifications/blob/master/source/objectid.rst). I used the same validation as the [bson package](https://github.com/mongodb/mongo-go-driver/blob/v1.9.1/bson/primitive/objectid.go#L78) uses.

I've also added en translation. I'm not fluent in any of the other languages. How do you usually add translations? I guess I can copy and paste together translations for objectid by looking at the other translations. Should I do that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ Baked-in Validations
 | jwt | JSON Web Token (JWT) |
 | latitude | Latitude |
 | longitude | Longitude |
+| objectid | ObjectID |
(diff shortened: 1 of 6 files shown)


zemzale:
This has been implemented with https://github.com/go-playground/validator/pull/1023 so I am closing this as resolved

PR #951, opened 2022-05-30
Expose the struct the field is contained within

wjam:
## Fixes Or Enhances

This allows users of `FieldError` to be able to query information about the struct the field is contained within. This information can then be used to work out what to display when the param field points to another field, such as with `gtfield`, and a tag name function has been registered.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -158,6 +158,9 @@ type FieldError interface {
 
 	// Error returns the FieldError's message
 	Error() string
+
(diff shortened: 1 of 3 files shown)


wjam:
The aim of this PR is to provide a method to solve #597.

Tnr1112:
Hi, I wonder when the solution will be in production. Or if there is a way to implement it in my project temporarily. Thanks!

zemzale (approved):

PR #975, opened 2022-07-07
Update errors.go

maltsev25:
**fix for aliases**
translate work incorrect for registered aliases

## Fixes Or Enhances

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -260,15 +260,19 @@ func (fe *fieldError) Error() string {
 // NOTE: if no registered translation can be found, it returns the original
 // untranslated error message.
 func (fe *fieldError) Translate(ut ut.Translator) string {
+	var fn TranslationFunc
 
 	m, ok := fe.v.transTagFunc[ut]
 	if !ok {
 		return fe.Error()
 	}
 
-	fn, ok := m[fe.tag]
+	fn, ok = m[fe.tag]
 	if !ok {
-		return fe.Error()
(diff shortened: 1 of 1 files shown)


maltsev25:
@deankarn Can you please review this PR?

errors.go
```diff
@@ -260,19 +260,22 @@ func (fe *fieldError) Error() string {
...
 	}
 
-	fn, ok := m[fe.tag]
+	fn, ok = m[fe.tag]
 	if !ok {
 		fn, ok = m[fe.actualTag]
 		if !ok {
 			return fe.Error()
 		}
-		return fe.Error()
```
hjwk, line 271: Looks like this is the only line that needs to change ?

hjwk (changes requested):
Can you squash your three commits into 1 ?

errors.go
maltsev25: You need to check variable `m` with 2 different tags `fe.tag` and `fe.actualTag`

hjwk: Ah, correct, got confused when looking at the diffs, with 1 commit it is clearer now

hjwk (approved):

maltsev25:
@deankarn Can you please review this PR?

maltsev25:
@deankarn Can you please review this PR?

deankarn (approved):
LGTM

I will try to merge and release this weekend, have a lot going on so may be a few days.

PR #979, opened 2022-07-28
add reflect struct field on validate error

ivybridge-3c33:
add reflect on fields 

use case: 
I've added reflect.StructField for using for getting a tag for using my own custom validate message and getting the field name from the language
example use from my own project

`
	if validate, ok := err.(*validator.InvalidValidationError); ok {
		panic(validate)
	}

	errorsFields := make(map[string][]string, 0)
	for _, validate := range err.(validator.ValidationErrors) {
		fe := validate.ReflectStructField()
		fieldName := strings.Replace(fe.Tag.Get("json"), ",omitempty", "", 0)
		if _, ok := errorsFields[fieldName]; !ok {
			errorsFields[fieldName] = make([]string, 0)
		}
		msg := ""
		switch errTagName := validate.Tag(); errTagName {
		case "regexp":
			msg = i18n.ParseT(ct, "validator.regexp") // invalid format blabla
		}
		errorsFields[fieldName] = append(errorsFields[fieldName], msg)
	}
`

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -75,11 +75,12 @@ type cStruct struct {
 }
 
 type cField struct {
-	idx        int
-	name       string
-	altName    string
-	namesEqual bool
-	cTags      *cTag
+	idx         int
+	name        string
+	altName     string
+	namesEqual  bool
+	cTags       *cTag
+	structField reflect.StructField
 }
 
 type cTag struct {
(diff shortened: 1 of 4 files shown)


deankarn:
@ivybridge-3c33 I like the idea of grabbing struct field, I wonder if it might be confusing though for when the validation is not a struct field.

I wonder if there would be a way for it to work with struct fields and regular ones also?

PR #982, opened 2022-07-30
feat: add image validation

the-fanan:
## Enhances

Make sure that you've checked the boxes below before you submit PR:
    ☑️ Tests exist or have been written that cover this particular change.

This PR does the following:
- adds a baked in validation to validate a file is a valid image
- adds a unit test to cover the added validation
- adds translations for image validation
- adds tests to cover translations for image validation
- updates the `README.md` document to include the added image validation

Motivation:
- Go is increasingly becoming used as a back-end language. File validations are pretty common for many back-end services. This validation will allow developers to easily validate that a file is a valid image.


the-fanan:
@hjwk

the-fanan:
@hjwk can you help review this please?

translations/ja/ja_test.go
```diff
@@ -139,6 +139,7 @@ func TestTranslations(t *testing.T) {
 		StrPtrGte         *string   `validate:"gte=10"`
 		OneOfString       string    `validate:"oneof=red green"`
 		OneOfInt          int       `validate:"oneof=5 63"`
+		Image			  string			`validate:"image"`
```
hjwk, line 142: type and tag and tag not aligned with lines above

hjwk (changes requested):
Sure, can do a quick review but I am not an owner...
Don't have much to say, but please check code alignment (I've highlighted one example in the review)

translations/ja/ja_test.go
the-fanan: Thank you. I'll work on that.

baked_in.go
```diff
@@ -1404,6 +1407,74 @@ func isFile(fl FieldLevel) bool {
...
+
+		if err != nil {
+			return false
+		}
+
+		if fileInfo.IsDir() {
+			return false
+		}
+
+		file, err := os.OpenFile(filePath, os.O_RDWR, 0644)
```
deankarn, line 1452: No need to have write permissions that I can see and `os.Open(...)` uses `O_RDONLY`
```suggestion
		file, err := os.Open(filePath)
```

baked_in.go
```diff
@@ -1404,6 +1407,74 @@ func isFile(fl FieldLevel) bool {
...
+
+		mime, err := mimetype.DetectReader(file)
+		if err != nil {
+			return false
+		}
+
+		_, err = file.Seek(0, io.SeekStart)
+		if err != nil {
+			panic(fmt.Sprintf("Could not reset seek start for file. Error: %s", err))
+		}
```
deankarn, line 1466: I'm not sure why the seek is required? should be removed from what I can tell.

translations/ar/ar_test.go
```diff
@@ -149,6 +149,7 @@ func TestTranslations(t *testing.T) {
 		PostCode          string            `validate:"postcode_iso3166_alpha2=SG"`
 		PostCodeCountry   string
 		PostCodeByField   string `validate:"postcode_iso3166_alpha2_field=PostCodeCountry"`
+		Image			  string			`validate:"image"`
```
deankarn, line 152: please run `gofmt` on all files changes 🙏

deankarn (commented):
A few changes and I think this PR would be good to go.

baked_in.go
the-fanan: While reading data on the mime type of the file, the offset for the reader is moved away from the beginning of the file and this may cause issues for other parts of the program that need to read from that file. Using seek ensures that after the mime type is read, the offset of the reader is returned to the beginning of the file so that there will be no issues for other parts of the program that will need to read from the file.

hiendaovinh:
Let's ship this please :+1:

baked_in.go
deankarn: Right I understand what `Seek` does and why one would need to use it but in this case the file handle is closed when this function ends because of the `defer file.Close()` and nothing else tries to read using that handle and so there is no need to Seek from what I can tell.

the-fanan: Right, got you.

deankarn (approved):

translations/pt_BR/pt_BR.go
```diff
@@ -1321,6 +1321,11 @@ func RegisterDefaultTranslations(v *validator.Validate, trans ut.Translator) (er
 			translation: "{0} deve ser um valor booleano válido",
 			override:    false,
 		},
+		{
+			tag:         "image",
+			translation: "{0} deve ser uma imagen válido",
```
eduardo-mior, line 1326: The correct translation is  
"{0} deve ser uma imagem válida"

Issue #974, opened 2022-07-07
hexadecimal validation in v10 validator


marcosartorato:
### Package version: 

v10

### Issue, Question or Enhancement:

An **hexadecimal** tag signals an invalid input over missing field even without the **required** tag

### Code sample, to showcase or reproduce:

Over a ConfUp struct w/ optional hexadecimal Fopts field and a `{}` input, the validation returns a non-nil err.

```go
type ConfUp struct{
	Fopts string `json:"fopts" validate:"hexadecimal"`
}
func (e *ConfUp) Validate() error {
	return validator.New().Struct(e)
}
func (e *ConfUp) FromByte(src []byte) error {
	return json.Unmarshal(src, e)
}
func (e *ConfUp) Byte() ([]byte, error) {
	return json.Marshal(e)
}
...
			cu := &ConfUp{}
			if ierr = cu.FromByte(payload); ierr == nil {
				if verr = cu.Validate(); verr == nil {
					return nil
				}
			}
...
```

I expected no check over a missing field for no **required** tag. Moreover, a **numeric** tag does not trigger the same behavior. Are there some docu coverage over this logic?

The **omitempty** tag bypasses the issue

Antonboom:
+1

PR #988, opened 2022-08-18
Fixed boolean validation to handle bool kind

V-R-Dighe:
## Fixes Or Enhances

**Make sure that you've checked the boxes below before you submit PR:**
- [x] Tests exist or have been written that cover this particular change.

@go-playground/validator-maintainers

PR #989, opened 2022-08-19
Fix: Hexadecimal validation in v10 validator

V-R-Dighe:
## Fixes Or Enhances
Resolves https://github.com/go-playground/validator/issues/974


deankarn:
This change appears to allow a blank string to pass as a valid hexidecimal value which is not correct.
The referenced issue shows that omitempty or the required tags are what you want preceeding this validation.

V-R-Dighe:
> This change appears to allow a blank string to pass as a valid hexidecimal value which is not correct. The referenced issue shows that omitempty or the required tags are what you want preceeding this validation.


Ok @deankarn , thanks for the reply.
Similar behaviour is there in most of the string tags. Do we always need to use an omitempty tag with them if the field is optional?
More simply, is the fix required for the [issue](https://github.com/go-playground/validator/issues/974)?

deankarn:
Hey @V-R-Dighe no a fix is not needed for that issue.

Yes an omitempty tag should be used for optional values when at least one other validation exists.

V-R-Dighe:
> Hey @V-R-Dighe no a fix is not needed for that issue.
> 
> Yes an omitempty tag should be used for optional values when at least one other validation exists.

Thanks @deankarn,  please review the one more PR which I raised [here](https://github.com/go-playground/validator/pull/988).

PR #998, opened 2022-09-14
add domain validation(simple domain names)

aryehlev:
## Fixes Or Enhances

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -122,6 +122,7 @@ var (
 		"email":                         isEmail,
 		"url":                           isURL,
 		"uri":                           isURI,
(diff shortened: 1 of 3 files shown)


deankarn:
@aryehlev I'm hesitant to merge this only because full domain validation is much more complicated than this and don't want to give users a false sense of security.

if a more comprehensive validation can be added I'd be happy to merge.

aryehlev:
This is the validation I use since I'd rather let some false negatives than false positives.
But I will try to research a more comprehensive validation and if I don't find one I will close this pr.
Thank you🙏

PR #1024, opened 2022-11-08
add validation for base64 url without = padding

### LINKED ISSUE #1002: base64url should not require padding
- [X] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [X] I have looked at the examples provided that may showcase my question [here](/_examples)?

v10

base64url validator is faulty. It requires padding even though padding is optional in URL safe encoding.

```
package main

import (
	"encoding/base64"
	"fmt"

)

func main() {
	msg := "hello"
	msgB64url := base64.RawURLEncoding.EncodeToString([]byte(msg))

	val := validator.New()

	fmt.Println(msgB64url) // prints "aGVsbG8"
	fmt.Println(val.Var(msgB64url, "base64url"))
}
```

yasoobh:
## Fixes Or Enhances

Change:

As flagged in [this](https://github.com/go-playground/validator/issues/1002) issue, validation fails for some strings encoded using base64.RawURLEncoding function.

The discussion around whether padding should be considered optional or not is rather long. Some of it can be found here: https://github.com/golang/go/issues/4237. Following which, RawURLEncoding was introduced in base64 package.

The general rule of thumb regarding padding seems to be that you MUST be padding when encoding, but SHOULD be able to handle both padded and unpadded inputs when decoding. Therefore, we should provide support for validating unpadded encoded inputs.

Instead of modifying the "base64url" tag to accept unpadded inputs, it makes more sense to have a separate tag - "base64rawurl". This way it would be clear to the developer and they would use one or the other, or both back to back, depending on their usecase.

PS: formatting changes in doc.go because of go 1.19 updates. Please lmk if I should revert them.

Closes #1002


azlancpool:
Will this change be added to any version soon?

apinet:
Any chance to have the last code review needed? 🥹

deankarn (approved):

deankarn (approved):

apinet:
@deankarn you are awesome! many thx for this gorgeous lib!

PR #1047, opened 2022-12-24
Added "http_url" tag to validate if the variable is a HTTP(s) URL

phith0n:
## Enhances

In most cases, the application only accepts the HTTP URL as input. But we don't have a suitable tag to validate it.

For example, application use `url` tag to validate the user input then use it as the cURL's parameter to download a network file:

```go
package main

import (
	"os"
	"os/exec"

)

func main() {
	var input = os.Args[1]
	var v = validator.New()
	err := v.Var(input, "url")
	if err != nil {
		panic(err)
	}

	cmd := exec.Command("curl", input)
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
	err = cmd.Run()
	if err != nil {
		panic(err)
	}
}
```

But it's not safe if user use `file:///etc/passwd` as the input, it will download the local file unexpected.

So I added a new tag named "http_url" to enhance the current situation.

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -121,6 +121,7 @@ var (
 		"e164":                          isE164,
 		"email":                         isEmail,
 		"url":                           isURL,
+		"http_url":                      isHttpURL,
 		"uri":                           isURI,
 		"urn_rfc2141":                   isUrnRFC2141, // RFC 2141
 		"file":                          isFile,
@@ -1365,6 +1366,24 @@ func isURL(fl FieldLevel) bool {
 		}
 
 		return true
+	}
+
+	panic(fmt.Sprintf("Bad field type %T", field.Interface()))
+}
+
+// isHttpURL is the validation function for validating if the current field's value is a valid HTTP(s) URL.
+func isHttpURL(fl FieldLevel) bool {
+	if !isURL(fl) {
+		return false
+	}
+
+	field := fl.Field()
+
+	switch field.Kind() {
+	case reflect.String:
+
+		s := field.String()
+		return strings.HasPrefix(s, "http://") || strings.HasPrefix(s, "https://")
 	}
 
 	panic(fmt.Sprintf("Bad field type %T", field.Interface()))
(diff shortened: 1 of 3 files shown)


baked_in.go
```diff
@@ -1370,6 +1371,23 @@ func isURL(fl FieldLevel) bool {
...
 
+// isHttpURL is the validation function for validating if the current field's value is a valid HTTP(s) URL.
+func isHttpURL(fl FieldLevel) bool {
+	if !isURL(fl) {
+		return false
+	}
+
+	field := fl.Field()
+	switch field.Kind() {
+	case reflect.String:
```
vijaynallagatla, line 1382: It'd be nice to see support for `*string` as thats the most used cases in most of the generators.

baked_in.go
```diff
@@ -1370,6 +1371,23 @@ func isURL(fl FieldLevel) bool {
...
+
+		s := strings.ToLower(field.String())
+		return strings.HasPrefix(s, "http://") || strings.HasPrefix(s, "https://")
```
vijaynallagatla, line 1385: How about making use of `url`?
```code
   u, err := url.Parse(s)
   if err != nil || u.Scheme == "http" || u.Scheme == "https" {
      panic(err)
   }
```

vijaynallagatla:
I liked this new TAG as it differs clearly from exiting `url` tag. However the tag name is not generic enough. `http_url` seems like it supports just the `http` scheme.

baked_in.go
deankarn: strings should be dereferenced by the time it reaches this point in the code.

baked_in.go
deankarn: I agree, except for the panic portion, that this would be fat superior to actually validate it's a valid URL which is far more than just the leading scheme.

deankarn (approved):
~Will merge but adjust to use `url.Parse`,~ scratch that will leave as is.

phith0n:
Hi @deankarn 

Thanks for your approval. Recently I'm so busy that couldn't response on time.

PR #1056, opened 2023-01-19
Include the field value in the error message

Schparky:
## Enhances


Schparky:
Decided against this since it may be a bad idea to include data in logs.

PR #939, opened 2022-05-04
Fix validator excluded_if

CorentinClabaut:
## Fixes Or Enhances
closes #938 

**Make sure that you've checked the boxes below before you submit PR:**
- [x] Tests exist or have been written that cover this particular change.

@go-playground/validator-maintainers

PR #1064, opened 2023-01-30
Fix validator `excluded_unless`

### LINKED ISSUE #1063: `excluded_unless` does not behave as expected

Relates to #938 

### Package version: 
 v10

Validator `excluded_unless` does not behave as the documentation describes. 
The documentation says: "The field under validation must not be present or is empty `unless` all the other specified fields are equal to the value following with the specified field."
What the validator does: "The field under validation must not be present or is empty `if` all the other specified fields are equal to the value following with the specified field."  

```go
type Testing struct {
	A string
	B string `validate:"excluded_unless=A include"`
}

errs = validate.Struct(Testing{A: "include", B: "pass"}) // -> Should pass, but fails
errs = validate.Struct(Testing{A: "exclude", B: "fail"}) // -> Should fail, but passes
```

jenoudet:
## Fixes Or Enhances
Fixes #1063 

@go-playground/validator-maintainers @deankarn

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -1623,10 +1623,10 @@ func excludedUnless(fl FieldLevel) bool {
 	}
 	for i := 0; i < len(params); i += 2 {
 		if !requireCheckFieldValue(fl, params[i], params[i+1], false) {
-			return true
+			return !hasValue(fl)
 		}
 	}
-	return !hasValue(fl)
+	return true
 }
 
 // excludedWith is the validation function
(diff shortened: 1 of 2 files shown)


smares:
Isn't this the same as https://github.com/go-playground/validator/pull/939?

jenoudet:
@smares No, #939 fixes `exluded_if` (do nothing, except if all conditions have been met, then exclude the field) , this PR fixes `excluded_unless` (exclude the field, except if all conditions have been met, then do nothing). They are opposite.

deankarn (approved):

PR #1106, opened 2023-05-15
feat:oneof support float whitout fractional

LjtGentle:
## Fixes Or Enhances

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/fs"
+	"math"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


deankarn (commented):
Changes look good, but curious, is there any reason why only allowing non fractional values?

LjtGentle:
It's my lack of consideration. My initial idea was to solve the problem of JSON serialization

baked_in.go
```diff
@@ -287,6 +290,13 @@ func isOneOf(fl FieldLevel) bool {
 	}
 	return false
 }
+func decimal2String(f float64) string {
+	_, frac := math.Modf(f)
+	if frac != 0 {
```
marlonbarreto-zulu, line 295: You can use this shorter syntax 🤗 

```go
	if _, frac := math.Modf(f); frac != 0 {
```

PR #1128, opened 2023-07-20
RegisterTagNameFunc doc update

teathedev:
## Fixes Or Enhances
RegisterTagNameFunc document was misleading. I updated the description to be more clear.
It doesn't affect the strings return from `StructField` so I added where it affect.

### DIFF
--- a/validator_instance.go
+++ b/validator_instance.go
@@ -190,7 +190,7 @@ func (v *Validate) ValidateMap(data map[string]interface{}, rules map[string]int
 	return v.ValidateMapCtx(context.Background(), data, rules)
 }
 
-// RegisterTagNameFunc registers a function to get alternate names for StructFields.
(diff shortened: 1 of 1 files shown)


validator_instance.go
```diff
@@ -190,7 +190,7 @@ func (v *Validate) ValidateMap(data map[string]interface{}, rules map[string]int
 	return v.ValidateMapCtx(context.Background(), data, rules)
 }
 
-// RegisterTagNameFunc registers a function to get alternate names for StructFields.
+// RegisterTagNameFunc registers a function to get alternate names (other than sturct field names) from FieldName.
```
deankarn, line 193: I don't think this is 100% correct either as if does not get them from `FieldName` but the `reflect.StructField`; perhaps it's still not specific enough?

PR #1145, opened 2023-08-16
add floating-point precision to four decimal places

ahaooahaz:
## Fixes Or Enhances

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -3,6 +3,7 @@ package validator
 import (
 	"context"
 	"fmt"
+	"math"
 	"reflect"
 	"strconv"
 )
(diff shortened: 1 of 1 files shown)


ahaooahaz:
@deankarn Floating-point numbers that lose precision should be a problem caused by reflection, I think the problem appears in `validator.go:77 current.Field(f.idx)`, if use
`current.Field(f.idx).Float()` it already lose precision ,I saw your code modify the 32-bit and 64-bit tool functions from string to float respectively, but this is not the actual cause of the problem

deankarn:
@ahaooahaz there is actually no problem per se.

This is a `floating point` precision thing and  because the `float32` is returned as a `float64` it's actuating gaining precision. 

Here's why my other PR fixes the issue without number truncation which can introduce other issues, especially where the code was implemented. 
The parsing code has been updated using `asFloat32` and `asFloat64` making it now behave identically to `reflect.Float()` function which both check the bit precision aka `float32` vs `float64`
```go
func asFloat64(param string) float64 {
	i, _ := strconv.ParseFloat(param, 64)
	return i
}

func asFloat32(param string) float64 {
	i, _ := strconv.ParseFloat(param, 32)
	return i
}

func main() {

	f32 := float32(0.1)
	f64 := float64(0.1)

	fmt.Println(float64(f32), reflect.ValueOf(f32).Float(), asFloat32("0.1"))
	fmt.Println(f64, reflect.ValueOf(f64).Float(), asFloat64("0.1"))
        //
	// output
	//
	// 0.10000000149011612 0.10000000149011612 0.10000000149011612
	// 0.1 0.1 0.1
}
```

ahaooahaz:
@deankarn ohhh, I get it, GREAT WORKS!!

Issue #1165, opened 2023-09-14
URL fails for file:///


pklanka:
- [x] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [x] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Code sample, to showcase or reproduce:

```go

package main

import (
	"fmt"

	"github.com/go-playground/validator/v10"
)

type SampleStruct struct {
	FileUrl string `json:"file_url" validate:"omitempty,url"`
}

func main() {

	Validator := validator.New()
	sample := SampleStruct{
		FileUrl: "file://tmp/abcd",
	}
	if err := Validator.Struct(sample); err != nil {
		// this passes
		fmt.Println(err)
	}

	sample2 := SampleStruct{
		FileUrl: "file:///tmp/abcd",
	}
	if err := Validator.Struct(sample2); err != nil {
		// this fails
		fmt.Println(err)
	}
}

```

pklanka:
file:///tmp/abcd should a valid file URL - ref: https://jkorpela.fi/fileurl.html

PR #1171, opened 2023-09-20
fix file URL validator

nodivbyzero:
## Fixes Or Enhances

This PR adds an additional check for the file URL schema.
Connected to #1165


deankarn (approved):
LGTM

PR #1185, opened 2023-10-26
refactor: use any instead of interface{}

chenyahui:
## Fixes Or Enhances


deankarn:
Sorry @chenyahui but this is a breaking change for anyone using Go 1.17 or below as you saw when needing to remove that from the CI test matrix to enable CI to pass.

I unfortunately cannot make such an update until cutting a new major version.

Issue #900, opened 2022-02-16
uuid4 doesn't works with google uuid.UUID


polRk:
- [+] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [+] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version eg. v9, v10: 

v10

### Issue, Question or Enhancement:

Got an error: `["3dcf8eb3-4489-4777-af93-2ba289316e77"] is not valid value for uuid.UUID`

### Code sample, to showcase or reproduce:

```go
type getContactRequest struct {
	ID uuid.UUID `uri:"id" binding:"required,uuid4"`
}
```

If i pass 

```go
type getContactRequest struct {
	ID string `uri:"id" binding:"required,uuid4"`
}
``` - It works

**Comment by zemzale** [comment]
Hey @polRk 

The validator works only for string types out of the box.

But you can register a CustomTypeFunc. It would look something like this: 
```go
package main

import (
	"fmt"
	"reflect"

	"github.com/go-playground/validator/v10"
	"github.com/google/uuid"
)

type getContactRequest struct {
	ID uuid.UUID `uri:"id" validate:"required,uuid4"`
}

// use a single instance of Validate, it caches struct info
var validate *validator.Validate

func main() {

	validate = validator.New()

	// register all sql.Null* types to use the ValidateValuer CustomTypeFunc
	validate.RegisterCustomTypeFunc(ValidateUUID, uuid.UUID{})

	// build object for validation
	x := getContactRequest{ID: uuid.New()}

	err := validate.Struct(x)

	if err != nil {
		fmt.Printf("Err(s):\n%+v\n", err)
	}
}

// ValidateUUID implements validator.CustomTypeFunc
func ValidateUUID(field reflect.Value) interface{} {
	if valuer, ok := field.Interface().(uuid.UUID); ok {
		return valuer.String()
	}
	return nil
}
```

See more at : https://github.com/go-playground/validator/blob/master/_examples/custom/main.go
and https://pkg.go.dev/github.com/go-playground/validator/v10#CustomTypeFunc and https://pkg.go.dev/github.com/go-playground/validator/v10#Validate.RegisterCustomTypeFunc

polRk:
Thank you!

polRk:
It doesn't work
```
["cb1cdadb-a93c-487b-be62-4963cef4600b"] is not valid value for uuid.UUID
```

```
[...]

PR #1189, opened 2023-10-30
Add support for validating against uuid values that are structs which implement the Stringer interface.

JoshGlazebrook:
## Fixes Or Enhances
This adds the ability to validate UUIDs that have an underlying struct value but also implement the Stringer interface. This should cover most UUID implementations (google's uuid, etc).

Implements: #900, specifically https://github.com/go-playground/validator/issues/900#issuecomment-1046030775.

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -508,47 +508,47 @@ func isASCII(fl FieldLevel) bool {
 
 // isUUID5 is the validation function for validating if the field's value is a valid v5 UUID.
 func isUUID5(fl FieldLevel) bool {
-	return uUID5Regex.MatchString(fl.Field().String())
+	return fieldMatchesRegexByStringerValOrString(uUID5Regex, fl)
 }
 
 // isUUID4 is the validation function for validating if the field's value is a valid v4 UUID.
 func isUUID4(fl FieldLevel) bool {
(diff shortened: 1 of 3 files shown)


deankarn:
@JoshGlazebrook this PR is close :) 

But I think it will fail for alias types that an implement `fmt.Stringer`. The type needs to be checked to see if the kind is already a string before attempting the cast. eg.

```go
type uuidAlias string

func (u uuidAlias) String() string {
	return "This is a UUID " + string(u)
}
```

I think the new helper can be changed like so to avoid this and maintain 100% backward compatibility:
```go
func fieldMatchesRegexByStringerValOrString(regex *regexp.Regexp, fl FieldLevel) bool {
	switch fl.Field().Kind() {
	case reflect.String:
		return regex.MatchString(fl.Field().String())
	default:
		if stringer, ok := fl.Field().Interface().(fmt.Stringer); ok {
			return regex.MatchString(stringer.String())
		} else {
			return regex.MatchString(fl.Field().String())
		}
	}
}
```

deankarn:
I will merge and then make these changes.

JoshGlazebrook:
Thank you!

PR #1196, opened 2023-11-15
MongoDB validator improved

pixel365:
## Enhances

Added MongoDB connection string validation


deankarn:
I'm not sure it would be best to have one validation do two different things, otherwise you wouldn't be able to differentiate for error messages and alike what exactly failed.

pixel365:
> I'm not sure it would be best to have one validation do two different things, otherwise you wouldn't be able to differentiate for error messages and alike what exactly failed.

I agree with this remark. Added a separate tag.

pixel365:
@deankarn Hi, how about this pull request?

deankarn (approved):

PR #1201, opened 2023-12-02
Fix Bad Field type panic

### LINKED ISSUE #907: Panic Bad field type *int64, when using required_if and omitempty at the same time
- [X] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [X] I have looked at the examples provided that may showcase my question [here](/_examples)?

v10.10.0

When validating an int64 pointer with `required_if` and `omitempty` at the same time, calling `Struct()` leads to panic.

Using `omitempty` merely is ok. Validating an int64 instead of an int64 pointer is ok.

Similar to: [Panic Bad field type *string, when using required_without and omitempty at the same time #654](https://github.com/go-playground/validator/issues/654)

```go
package main

import (
	"log"
)

type Sample struct {
	Type     string `json:"type" validate:"required"`
	Quantity int64  `json:"quantity" validate:"required_if=Type special,gte=2"`
}

type SampleWithPointer struct {
	Type     string `json:"type" validate:"required"`
	Quantity *int64 `json:"quantity,omitempty" validate:"required_if=Type special,gte=2"`
}

func main() {

	_type := "notspecial"
	s := &Sample{Type: _type}
	swp := &SampleWithPointer{Type: _type}

	validate := validator.New()
	if err := validate.Struct(s); err != nil {
		log.Printf("Validate sample with error, %v", err)
	}
	if err := validate.Struct(swp); err != nil {
		log.Printf("Validate SampeWithPointer with error, %v", err)
	}
}
```
stack trace:
```
2009/11/10 23:00:00 Validate sample with error, Key: 'Sample.Quantity' Error:Field validation for 'Quantity' failed on the 'gte' tag
panic: Bad field type *int64

goroutine 1 [running]:
github.com/go-playground/validator/v10.isGte({0x59b798, 0xc0000d1c20})
	/tmp/gopath931800745/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/baked_in.go:1789 +0x4dd
github.com/go-playground/validator/v10.wrapFunc.func1({0x534940, 0xc0000ac208}, {0x59b798, 0xc0000d1c20})
	/tmp/gopath931800745/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/baked_in.go:40 +0x2d
github.com/go-playground/validator/v10.(*validate).traverseField(0xc0000d1c20, {0x599a70, 0xc0000ba000}, {0x5471a0, 0xc0000ac1f8, 0x13}, {0x534940, 0xc0000ac208, 0x85}, {0xc000350300, ...}, ...)
	/tmp/gopath931800745/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/validator.go:445 +0x16a7
github.com/go-playground/validator/v10.(*validate).validateStruct(0xc0000d1c20, {0x599a70, 0xc0000ba000}, {0x534e00, 0xc0000ac1f8, 0x1}, {0x5471a0, 0xc0000ac1f8, 0xc0000a31e0}, {0x59bd60, ...}, ...)
	/tmp/gopath931800745/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/validator.go:77 +0x765
github.com/go-playground/validator/v10.(*Validate).StructCtx(0xc0003492c0, {0x599a70, 0xc0000ba000}, {0x534e00, 0xc0000ac1f8})
	/tmp/gopath931800745/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/validator_instance.go:344 +0x430
github.com/go-playground/validator/v10.(*Validate).Struct(...)
	/tmp/gopath931800745/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/validator_instance.go:317
main.main()
	/tmp/sandbox1464673145/prog.go:29 +0x13d

Program exited.
```

darklam:
## Fixes Or Enhances
Fixes #907 (and possibly others?)

At first I found it weird that the `required_if` affected the behavior of the `gte` tag, but after some debugging it seems that the condition `runValidationWhenNil` is not checked for each field tag, but only for the first one. 
In this case it was set to true for the `required_if` tag, but set to false for `gte`, so when the `required_if` preceded the `gte` tag it meant that the field was incorrectly validated through `gte` even when it was nil. 

### DIFF
--- a/validator.go
+++ b/validator.go
@@ -445,6 +445,27 @@ OUTER:
 			}
 
 		default:
+			switch current.Kind() {
+			case reflect.Ptr, reflect.Interface:
+				if !ct.runValidationWhenNil && current.IsNil() {
+					v.errs = append(v.errs,
+						&fieldError{
+							v:              v.v,
+							tag:            ct.aliasTag,
+							actualTag:      ct.tag,
+							ns:             v.str1,
+							structNs:       v.str2,
+							fieldLen:       uint8(len(cf.altName)),
+							structfieldLen: uint8(len(cf.name)),
+							value:          current.Interface(),
+							param:          ct.param,
+							kind:           kind,
+							typ:            current.Type(),
+						},
+					)
+					return
+				}
+			}
 
 			// set Field Level fields
 			v.slflParent = parent
--- a/validator_test.go
+++ b/validator_test.go
@@ -697,6 +697,19 @@ func TestAliasTags(t *testing.T) {
 	PanicMatches(t, func() { validate.RegisterAlias("exists!", "gt=5,lt=10") }, "Alias 'exists!' either contains restricted characters or is the same as a restricted tag needed for normal operation")
 }
 
+func TestNoPanicOnRequiredIfBeforeGTEWithPointer(t *testing.T) {
+	val := New()
+
+	type WithPointer struct {
+		Type     string `validate:"required"`
+		Quantity *int64 `validate:"required_if=Type special,gte=2"`
+	}
+	structWithPointer := &WithPointer{Quantity: nil, Type: "notspecial"}
+
+	errs := val.Struct(structWithPointer)
+	NotEqual(t, errs, nil)
+}
+
 func TestNilValidator(t *testing.T) {
 	type TestStruct struct {
 		Test string `validate:"required"`


validator.go
```diff
@@ -445,6 +445,27 @@ OUTER:
 			}
 
 		default:
+			switch current.Kind() {
+			case reflect.Ptr, reflect.Interface:
+				if !ct.runValidationWhenNil && current.IsNil() {
+					v.errs = append(v.errs,
+						&fieldError{
```
darklam, line 452: Not very familiar with the errors in this library, so I just went by what was used in the other instance where the `runValidationWhenNil` is checked and could be entirely wrong 😄

deankarn:
Thanks fro the PR @darklam 

I will have to think hard on if this could cause any breaking side-effects, see [here](https://github.com/go-playground/validator/issues/907#issuecomment-1849000624) for details.

darklam:
> Thanks fro the PR @darklam
> 
> I will have to think hard on if this could cause any breaking side-effects, see [here](https://github.com/go-playground/validator/issues/907#issuecomment-1849000624) for details.

Yes, agreed that this should be scrutinized a lot regarding breaking changes before merging. I'm happy to add more test cases if there are any specific ones coming to mind.
Also, my first thought was indeed to change some baked ins like gte to handle nil cases, but it didn't seem like the way to go because it became quite repetitive.
I think that the check for that should happen before getting to these validators (as the original intention seems to be with the `runValidationWhenNil` flag), but of course this is not an informed opinion as I'm quite new with the codebase 😄

PR #1211, opened 2024-01-10
feat: omit anonymous embedded struct

### LINKED ISSUE #776: Method Struct(...) forms keys for embedded structures incorrectly

v10

The validator sees no difference between an additional field of type "struct" and embedded structures. This complicates further post-processing and forces you to pre-validate each structure and then merge the results.

Is there any way to correct this behavior? For example, by adding an additional tag.

```go
package main

import (
	"fmt"

)

type User struct {
	Name string `validate:"required"`
}

type GroupA struct {
	Admin User
}

type GroupB struct {
	User
}

func main() {
	v := validator.New()

	errs := v.Struct(new(GroupA))
	fmt.Println("GroupA: ", errs)

	errs = v.Struct(new(GroupB))
	fmt.Println("GroupB: ", errs)
}
```
#### Output
```
GroupA:  Key: 'GroupA.Admin.Name' Error:Field validation for 'Name' failed on the 'required' tag
GroupB:  Key: 'GroupB.User.Name' Error:Field validation for 'Name' failed on the 'required' tag
```
But for "GroupB" the key "GroupB.Name" is expected.

abemedia:
## Fixes Or Enhances

This PR adds an option to omit the names of anonymous embedded structs.

For example right now the code below would produce a namespace of `Foo.Bar.Baz`, with this code the option `WithOmitAnonymousName` would instead display `Foo.Baz`.

```go
type Foo struct {
  Bar
}

type Bar struct {
  Baz string `validate:"required"`
}
```

- Fixes #776

I'm not that familiar with the inner workings of `validator` so apologies if this is implemented in a terrible way. Happy to change it if the concept is acceptable and I get some pointers.

### DIFF
--- a/options.go
+++ b/options.go
@@ -14,3 +14,10 @@ func WithRequiredStructEnabled() Option {
 		v.requiredStructEnabled = true
 	}
 }
+
+// WithRequiredStructEnabled enables omitting the name of embedded anonymous structs from the namespace.
+func WithOmitAnonymousName() Option {
+	return func(v *Validate) {
+		v.omitAnonymousName = true
+	}
+}
--- a/validator.go
+++ b/validator.go
@@ -31,7 +31,6 @@ type validate struct {
 
 // parent and current will be the same the first run of validateStruct
 func (v *validate) validateStruct(ctx context.Context, parent reflect.Value, current reflect.Value, typ reflect.Type, ns []byte, structNs []byte, ct *cTag) {
-
 	cs, ok := v.v.structCache.Get(typ)
 	if !ok {
 		cs = v.v.extractStructCache(current, typ.Name())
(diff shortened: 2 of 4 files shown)


abemedia:
@deankarn any chance of getting this in?
If the approach using an option is not preferable I'm also happy to implement an alternative approach, for example having an `inline` tag the way the stdlib does for JSON.

zemzale (commented):
Very late on reviewing this, but this seems like a good fix for this issue and also can be used for fixing https://github.com/go-playground/validator/issues/1413 and I would like to merge this over https://github.com/go-playground/validator/pull/1426

@deankarn What do you think? 

@abemedia If you could resolve the conflict and fix the copy paste issue I can approve this.

deankarn:
@zemzale I think more discussion is required.

iirc I think there is more than this one place that might need this check when looking into this exiting PR for the same functionality https://github.com/go-playground/validator/pull/1426

See the comments on that other PR for details but I was hoping we could fix this with the RegisterTagNameFunc instead. The original idea was when returning blank from there it would omit the current namespace item, however it looks like we’ve had a bug with it for a long time and not sure how to correct in a backward compatible way yet. Also been busy with work and family and why I haven’t been able to look more into it.

abemedia:
Closed in favour of
- #1567

Issue #943, opened 2022-05-13
required_without rule results in panic or error when validateMap is used


benjamin-eich-publicare:
- [x] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [x] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version eg. v9, v10:
v10 (v10.11.0)

### Issue, Question or Enhancement:

There seems to be a problem with the required_without validation. It might also affect other rules that validate a field against other fields.

Please see the code examples below for expected and observed behavior.

If more information is needed I'm happy to provide it. Thanks for your help!

#### Example 1
```go
v := validator.New()

dataMap := map[string]interface{}{
    "foo": "bar",
    "foo2": "",
}
validationRules := map[string]interface{}{
    "foo": "required_without=foo2",
}
res := v.ValidateMap(dataMap, validationRules)
```
results in
```
panic: Invalid field namespace
```
Expected result would be an empty res because everything is fine.

#### Example 2
```go
v := validator.New()

dataMap := map[string]interface{}{
//    "foo": "",
    "foo2": "bar",
}
validationRules := map[string]interface{}{
    "foo": "required_without=foo2",
}
res := v.ValidateMap(dataMap, validationRules)

fmt.Printf("%+v\n", res)
```
results in (no difference if foo is provided or not)
```
map[foo:Key: '' Error:Field validation for '' failed on the 'required_without' tag]
```
Expected result would be an empty res because everything is fine.

Issue #893, opened 2022-02-06
required_if panics with ValidateMap on simple map with a single rule


dosco:
- [ X] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [ X] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version eg. v9, v10: 

v10

### Issue, Question or Enhancement:

required_if panics `panic("Invalid field namespace")` when using ValidateMap with simple rules.

### Code sample, to showcase or reproduce:

https://go.dev/play/p/6do5LR2UroT

```go

package main

import (
	"fmt"

	"github.com/go-playground/validator/v10"
)

func main() {
	data := map[string]interface{}{
		"name": "hello",
		"id":   123,
	}

	rules := map[string]interface{}{
		"name": "required_if=id 345",
		"id":   123,
	}

	v := validator.New()
	fmt.Println(v.ValidateMap(data, rules))
}

```

```
panic: Invalid field namespace

goroutine 1 [running]:
github.com/go-playground/validator/v10.(*validate).getStructFieldOKInternal(0xe55630636d8df4e, {0x538880, 0x595648, 0xc00029f670}, {0x5672a7, 0xc00030a6e0})
	/tmp/gopath1590948958/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/util.go:220 +0x160b
github.com/go-playground/validator/v10.(*validate).GetStructFieldOKAdvanced2(0x6bc240, {0x538880, 0x595648, 0x6}, {0x5672a7, 0x0})
	/tmp/gopath1590948958/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/field_level.go:119 +0x28
github.com/go-playground/validator/v10.requireCheckFieldValue({0x59a318, 0xc00007b9e0}, {0x5672a7, 0x2}, {0x5672aa, 0x3}, 0x8)
	/tmp/gopath1590948958/pkg/mod/github.com/
```

cchengubnt:
this issue has been a long time, any plan fix it and merge in to main branch?

zein-adi:
+1 i used version 10.15.5 the problem still exists

thedenisnikulin:
this is weird... How such a long maintained and popular package has such a bug which hasn't been fixed since 2022? Isn't "required_if" an essential tag?

Issue #876, opened 2022-01-10
ValidateMap check gtefield panic


f4haofeng:
when i run test case like this，
```go
func TestValidatorMap(t *testing.T) {
	validate := validator.New()
	data := map[string]interface{}{"start": 1, "end": 2}
	rules := map[string]interface{}{"end": "gtefield=start"}

	errs := validate.ValidateMap(data, rules)

	for k, err := range errs {
		fmt.Println(k)
		fmt.Println(err)
	}
}
```
output the message of panic：

panic: Invalid field namespace [recovered]
	panic: Invalid field namespace

goroutine 33 [running]:
testing.tRunner.func1.1(0x59d4040, 0x6037490)
	/Users/zhenghaofeng/go1.5/src/testing/testing.go:1072 +0x30d
testing.tRunner.func1(0xc000683e00)
	/Users/zhenghaofeng/go1.5/src/testing/testing.go:1075 +0x41a
panic(0x59d4040, 0x6037490)
	/Users/zhenghaofeng/go1.5/src/runtime/panic.go:969 +0x1b9
github.com/go-playground/validator/v10.(*validate).getStructFieldOKInternal(0xc0001338c0, 0x59d33c0, 0x602cbe8, 0x82, 0x5d9e405, 0x5, 0x8, 0x203000, 0x5, 0x203000, ...)
	/Users/zhenghaofeng/go/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/util.go:220 +0x1d35
github.com/go-playground/validator/v10.(*validate).GetStructFieldOK(0xc0001338c0, 0x59d33c0, 0x602cbe8, 0x82, 0x5d835fa, 0xc000351728)
	/Users/zhenghaofeng/go/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/field_level.go:98 +0x69
github.com/go-playground/validator/v10.isGteField(0x60dc5e0, 0xc0001338c0, 0x40104d8)
	/Users/zhenghaofeng/go/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/baked_in.go:1657 +0x89
github.com/go-playground/validator/v10.wrapFunc.func1(0x60b0780, 0xc0001a6008, 0x60dc5e0, 0xc0001338c0, 0x59d33c0)
	/Users/zhenghaofeng/go/pkg/mod/github.com/go-playground/validator/v10@v10.10.0/baked_in.go:40 +0x39
[...]

PR #1219, opened 2024-02-01
fix ValidateMap and field references

jippi:
## Fixes Or Enhances

Implement support for `ValidateMap` using `Field` references 

Fixes https://github.com/go-playground/validator/issues/893
Fixes https://github.com/go-playground/validator/issues/876
Fixes https://github.com/go-playground/validator/issues/943


validator_instance.go
```diff
@@ -162,17 +162,22 @@ func (v *Validate) SetTagName(name string) {
 // ValidateMapCtx validates a map using a map of validation rules and allows passing of contextual
 // validation information via context.Context.
 func (v Validate) ValidateMapCtx(ctx context.Context, data map[string]interface{}, rules map[string]interface{}) map[string]interface{} {
+	return v.validateMapCtx(ctx, data, data, rules)
+}
+
+// validateMapCtx will track the original "root" map (in case of nesting) which will allow for Field reference validation
```
jippi, line 168: I was not sure how y'all prefer naming this, so I just lowercased the "inner loop" func here

validator_instance.go
```diff
@@ -656,6 +661,34 @@ func (v *Validate) VarCtx(ctx context.Context, field interface{}, tag string) (e
...
+		return nil
+	}
+
+	ctag := v.fetchCacheTag(tag)
+
+	val := reflect.ValueOf(field)
+	vd := v.pool.Get().(*validate)
+	vd.top = val
+	vd.isPartial = false
+	vd.traverseField(ctx, reflect.ValueOf(parent), val, vd.ns[0:0], vd.actualNs[0:0], defaultCField, ctag)
```
jippi, line 682: this is the only difference from `VarCtx` (2nd arg of `reflect.ValueOf(parent)`) - I found it cleaner to have a new func for it, but fine with any other feedback to this y'all have

jippi:
bump on this one 🙏

jippi:
bumping on this @deankarn

util.go
```diff
@@ -212,7 +215,13 @@ BEGIN:
 		// reflect.Type = string
 		default:
 			val = current.MapIndex(reflect.ValueOf(key))
-			namespace = namespace[endIdx+1:]
+			// If we exceeded the length of the namespace, then we're at the end
+			// of the namespace traversal, so set it to empty value to break loop
+			if endIdx+1 > len(namespace) {
```
ccoVeille, line 220: Is this code formatted ?

I would expect


```suggestion
			if endIdx + 1 > len(namespace) {
```

Maybe I'm wrong, but a gofmt might help

validator_instance.go
```diff
@@ -656,6 +661,35 @@ func (v *Validate) VarCtx(ctx context.Context, field interface{}, tag string) (e
...
+// validate Array, Slice and maps fields which may contain more than one error
+func (v *Validate) VarCtxMap(ctx context.Context, key string, parent interface{}, field interface{}, tag string) (err error) {
+	if len(tag) == 0 || tag == skipValidationTag {
+		return nil
+	}
+
+	ctag := v.fetchCacheTag(tag)
+
+	val := reflect.ValueOf(field)
+	vd := v.pool.Get().(*validate)
```
ccoVeille, line 679: No type assertion check here?


```suggestion
	vd, ok := v.pool.Get().(*validate)
  if !ok {
       return nil
  }
```

PR #1224, opened 2024-02-12
feat: URN RFC 8141

### LINKED ISSUE #1223: URN (RFC 8141) validator

v10

Enable validation of a string value as a URN according to [RFC 8141](https://datatracker.ietf.org/doc/html/rfc8141) spec (2017).

```go
type Config struct {
    path string `validate:"urn"`
}
```

leodido:
## Fixes Or Enhances

Fixes #1223

Follow-up from #429 by adding also the most recent URN validator as per RFC 8141 spec.


baked_in.go
```diff
@@ -130,6 +130,7 @@ var (
 		"url":                           isURL,
 		"http_url":                      isHttpURL,
 		"uri":                           isURI,
+		"urn":                           isUrnRFC8141, // RFC 8141
```
deankarn, line 133: One requested change to match the rfc
```suggestion
		"urn_rfc8141":                           isUrnRFC8141, // RFC 8141
```

nodivbyzero:
LGTM

@leodido  Could you update the README.md with `urn_rfc8141`?

nodivbyzero (approved):

leodido:
Ping

baked_in.go
```diff
@@ -1505,6 +1506,23 @@ func isHttpURL(fl FieldLevel) bool {
...
 }
 
+// isUrnRFC8141 is the validation function for validating if the current field's value is a valid URN as per RFC 8141.
+func isUrnRFC8141(fl FieldLevel) bool {
+	field := fl.Field()
+
+	switch field.Kind() {
+	case reflect.String:
+
+		str := field.String()
```
nodivbyzero, line 1706: If str == "", we should return false.

nodivbyzero (changes requested):

nodivbyzero:
@leodido 
I reviewed this PR again. One minor issue and it's good to go.

baked_in.go
leodido: Empty input is already rejected by go-urn in RFC 8141 mode, but making that contract explicit here is reasonable. I’ll add the early return and a corresponding test case.

leodido:
@nodivbyzero done! PTAL

nodivbyzero (approved):

PR #1231, opened 2024-02-14
Make `url` and `http_url` work with url.URL.

bfabio:
## Fixes Or Enhances

This makes the `url` and `http_url` validations work with `url.URL` or any newtype that implements the `Stringer` interface. I'm not familiar with the codebase so it's kind of a PoC PR to see if you think this can be useful.

But they don't pass, if this is a feature we want and it's the right approach I can investigate deeper. I suspect the reflection magic isn't working as I expected.

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -1453,33 +1453,37 @@ func isFileURL(path string) bool {
 // isURL is the validation function for validating if the current field's value is a valid URL.
 func isURL(fl FieldLevel) bool {
 	field := fl.Field()
+	var s string
 
 	switch field.Kind() {
 	case reflect.String:
-
-		s := strings.ToLower(field.String())
-
-		if len(s) == 0 {
-			return false
+		s = strings.ToLower(field.String())
+	default:
+		if stringer, ok := field.Interface().(fmt.Stringer); ok {
+			s = stringer.String()
+		} else {
+			panic(fmt.Sprintf("Bad field type %T", field.Interface()))
 		}
+	}
 
-		if isFileURL(s) {
-			return true
-		}
+	if len(s) == 0 {
+		return false
+	}
 
-		url, err := url.Parse(s)
-		if err != nil || url.Scheme == "" {
-			return false
-		}
-
-		if url.Host == "" && url.Fragment == "" && url.Opaque == "" {
-			return false
-		}
-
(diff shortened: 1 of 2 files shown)


baked_in.go
```diff
@@ -1453,33 +1453,37 @@ func isFileURL(path string) bool {
...
-
-		if len(s) == 0 {
-			return false
+		s = strings.ToLower(field.String())
+	default:
+		if stringer, ok := field.Interface().(fmt.Stringer); ok {
+			s = stringer.String()
+		} else {
+			panic(fmt.Sprintf("Bad field type %T", field.Interface()))
 		}
```
nodivbyzero, line 1466:
```suggestion
		if stringer, ok := field.Interface().(fmt.Stringer); !ok {
		        panic(fmt.Sprintf("Bad field type %T", field.Interface()))
		}
		s = stringer.String()

```

baked_in.go
```diff
@@ -1453,33 +1453,37 @@ func isFileURL(path string) bool {
...
-		}
+	if isFileURL(s) {
+		return true
+	}
 
-		if url.Host == "" && url.Fragment == "" && url.Opaque == "" {
-			return false
-		}
+	url, err := url.Parse(s)
+	if err != nil || url.Scheme == "" {
```
nodivbyzero, line 1478:
```suggestion
	if err != nil {
```

baked_in.go
```diff
@@ -1453,33 +1453,37 @@ func isFileURL(path string) bool {
...
+		return false
+	}
 
-		return true
+	if url.Host == "" && url.Fragment == "" && url.Opaque == "" {
```
nodivbyzero, line 1482:
```suggestion
	if url.Scheme == "" || (url.Host == "" && url.Fragment == "" && url.Opaque == "") {
```

nodivbyzero:
I've added a few comments suggesting readability improvements.
I recommend keeping the original test functions unchanged and adding new ones for the `Stringer` type. Also, ensure the GitHub Actions checks pass.

bfabio:
I rebased to include #1444 and tweaked it a bit.

@nodivbyzero @go-playground/validator-maintainers

bfabio:
@nodivbyzero wdyt?

Issue #417, opened 2018-11-14
Validating unexported fields


Jwata:
### Package version eg. v8, v9: 
v8 and v9

### Question:
I was wondering if it's possible to validate unexported fields. But unexported fields are skipped here.
https://github.com/go-playground/validator/blob/v9/cache.go#L131

Is there any reason for it?

joeybloggs:
@Jwata A package can only view the unexported fields of types within its own package and so they must be ignored or there will be an exception; this is true of all packages using reflect even the son package.

Jwata:
OK, so they must be exported to do validate using reflection.

fuintis:
As far as i know reflect can read an unexported field, but write can't. Check the following link: https://yourbasic.org/golang/access-private-field-reflection/. Also you can use reflect2 package as json-iterator/extra do: https://godoc.org/github.com/json-iterator/go/extra#SupportPrivateFields.

Is there's a way that you considirer validate unexported fields?

Many thanks un advance.

deankarn:
@fuintis thanks for the link, I could have sworn this was not possible but I’ll take another look! love to support unexported fields if I can

bmassemin:
Should this issue be reopened ?
It could be handy to have a validation on unexported fields when you want your struct to be immutable.

sz-po:
Same here. This issue completely breaks DDD entity pattern.

gl19:
Also have the same feeling. Would love this to be looked into.

Antonboom:
+1

hampgoodwin:
Yeah,
This would be fantastic if it's possible 🙏

dixonwhitmire:
Just chiming in. I ran into this issue today. I am new to Go, but happy to help develop, test, or review a proposed solution!

dixonwhitmire:
> Just chiming in. I ran into this issue today. I am new to Go, but happy to help develop, test, or review a proposed solution!

Hello @Jwata and @deankarn ! I just saw that this issue is re-opened. Is there anything I can do to help?

Jwata:
I reopened this thread because other people also want this feature. 

> Is there anything I can do to help?

It seems @fuintis gave some ideas, which you may want to check.

PR #1234, opened 2024-02-22
Resolving "Validating unexported fields #417"

nikolaianohyn:
## Add supporting for validation private fields https://github.com/go-playground/validator/issues/417


mgerasimchuk (approved):

deankarn (changes requested):
Thanks for the PR!

I still have to review carefully, new laptop still in the mail so be a bit, but due to the use of unsafe I will request that this needs to become an opt-in option when initializing validator and documentation can indicate that and that opting in accepts that risk.

nikolaianohyn:
@deankarn thanks for your proposal, it looks valid. Changes added.

nikolaianohyn:
@deankarn if I can speed up the review process, let me know.

deankarn (approved):
LGTM

I will add a little more to the opt-in function comment but code looks great!

PR #1246, opened 2024-04-01
Lazily initialize regexes to speed up `init()`

kylecarbs:
Using `GODEBUG=inittrace=1` I found a relatively constant 10ms init caused by this function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Package validator
 =================
 <img align="right" src="logo.png">[![Join the chat at https://gitter.im/go-playground/validator](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/go-playground/validator?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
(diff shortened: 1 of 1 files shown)


deankarn:
@kylecarbs out of curiosity, what is your use case where `10ms` is not tolerable?

kylecarbs:
@deankarn this package is imported in our CLI. Adding that overhead to every invocation (even help) isn't great.

kylecarbs:
@deankarn any changes needed for me to push this through?

deankarn:
@kylecarbs sorry for the lateness of my reply, been busy.

My only concern is the `sync.Mutex` overhead for longer running applications, can you change the implementation to use [sync.Once](https://pkg.go.dev/sync#Once) instead which uses lower level atomics?

kylecarbs:
@deankarn good idea. Will change in a bit!

deankarn:
I will check back later tonight/tomorrow :)

Issue #1266, opened 2024-05-25
Proposal: Support validator chaining


mxmauro:
- [ X ] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [ X ] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version:

v10

### Enhancement:

Current implementation does not allow to augment predefined validators.

Let's say someone define a custom type and want to add support for `isEq`, `isNe`, etc., If he/she calls `RegisterValidation`, the existing validator will be overriden completely and existing functionality for them must be written again.

### Proposal.

To avoid the library to keep a chain of validators and break compatibility adding a new return value that indicates if the callback handled the validator or the library must call the next one in the chain, I simply propose adding a `GetRegisterValidation` routine so the developer can manage chaining.

If the idea is accepted, I can create a PR.

Regards,
Mauro.

PR #1267, opened 2024-05-25
Validation chain

mxmauro:
## Enhancement

Please refer to: https://github.com/go-playground/validator/issues/1266

- [ X ] Tests exist or have been written that cover this particular change.


mxmauro:
Unable to get the reason why the coverage goes down because coverage details gives status 500.

PR #1279, opened 2024-06-13
Use sync.OnceValue for lazy regexp initialization

kke:
## Fixes Or Enhances

Using sync.OnceValue instead of the local implementation brings some benefits in allocations, panic handling and mutex usage after initialization.


kke:
> coverage: 74.285% (-0.006%) from 74.291%

The coverage check does not seem to correctly calculate when files with build tags are involved.

PR #1282, opened 2024-06-17
feat: add validator for IPv4 bind address / host:port

rtomadpg:
## Fixes Or Enhances

Here is a little PR to say thank you, for providing your framework.

Let's add a validator for IPv4 bind addresses.

Examples of valid input:

- `:9000` for port 9000 on any IP address
- `1.2.3.4:7777` for port 7777 on specific IP address

My use case: I need this validator for an application's config including IP based socket listener addresses, like: `:9999` or `127.0.0.1:9999`


baked_in.go
```diff
@@ -2704,6 +2705,29 @@ func isHostnamePort(fl FieldLevel) bool {
...
+		return false
+	}
+
+	// If IP address is specified, it should match a valid IPv4 address
+	if ip != "" {
+		// we need to support older Golang versions, so we can not use netip.ParseAddr
+		parsedIp := net.ParseIP(ip)
+		return parsedIp != nil && parsedIp.To4() != nil
+	}
+	return true
```
nodivbyzero, line 2728: More readable and concise implementation:
```suggestion
	host, port, err := net.SplitHostPort(val)
	if err != nil {
		return false
	}
	// Validate port range (1-65535).
	if portNum, err := strconv.ParseUInt(port, 10, 32,); err != nil || portNum < 1 || portNum > 65535 {
		return false
	}

	// If no host is specified, return true (valid)
        if host == "" {
            return true
        }

        // Parse and validate IPv4 address
        ip := net.ParseIP(host)
        return ip != nil && ip.To4() != nil
```

baked_in.go
```diff
@@ -2704,6 +2705,28 @@ func isHostnamePort(fl FieldLevel) bool {
...
 
+// isIPv4Port validates a <ipv4>:<port> combination for fields typically used for socket address.
+func isIPv4Port(fl FieldLevel) bool {
+	val := fl.Field().String()
+	host, port, err := net.SplitHostPort(val)
+	if err != nil {
+		return false
+	}
+	// Validate port range (1-65535).
+	if portNum, err := strconv.ParseUInt(port, 10, 32,); err != nil || portNum < 1 || portNum > 65535 {
```
alob-mtc, line 2716:
```suggestion
	if portNum, err := strconv.ParseUint(port, 10, 32); err != nil || portNum < 1 || portNum > 65535 {
```

Issue #1268, opened 2024-05-26
Multiple validations at one time


reversersed:
- [x] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [x] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version eg. v9, v10: 

v10

### Issue, Question or Enhancement:

I have several validation tags, but validator works sequently, so I get only one error per field at the time. Is there some way to validate all tags at one time?

### Code sample, to showcase or reproduce:

```go
type UserRegisterQuery struct {
	Login    string `json:"login" validate:"required,min=4,max=16" example:"user"`
	Email    string `json:"email" validate:"required,email" example:"user@example.com"`
	Password string `json:"password" validate:"required,min=8,max=32,lowercase,uppercase,digitrequired,specialsymbol" example:"User!1password"`
}

func ValidationError(errors validator.ValidationErrors, dev_message string) *Error {
	var errs []string
	for _, err := range errors {
		switch err.Tag() {
		case "required":
			errs = append(errs, fmt.Sprintf("%s: field is required", err.Field()))
		case "oneof":
			errs = append(errs, fmt.Sprintf("%s: field can only be: %s", err.Field(), err.ActualTag()))
		case "min":
			errs = append(errs, fmt.Sprintf("%s must be at least %s characters length", err.Field(), err.ActualTag()))
		case "max":
			errs = append(errs, fmt.Sprintf("%s can't be more that %s characters length", err.Field(), err.ActualTag()))
		case "email":
			errs = append(errs, fmt.Sprintf("%s must be a valid email", err.Field()))
		case "jwt":
			errs = append(errs, fmt.Sprintf("%s must be a JWT token", err.Field()))
		case "lowercase":
			errs = append(errs, fmt.Sprintf("%s must contain at least one lowercase character", err.Field()))
		case "uppercase":
			errs = append(errs, fmt.Sprintf("%s must contain at least one uppercase character", err.Field()))
		case "digitrequired":
			errs = append(errs, fmt.Sprintf("%s must contain at least one digit", err.Field()))
		case "specialsymbol":
			errs = append(errs, fmt.Sprintf("%s must contain at least one special symbol", err.Field()))
		default:
			errs = append(errs, err.Error())
		}
	}
	return NewError(errs, "IE-0004", dev_message)
}

func Handler() error {
	...
	if err := h.Validator.Struct(query); err != nil {
[...]

PR #1288, opened 2024-07-03
Change to allow multiple errors be returned from single field when option "WithMultipleErrorsReturned" is set.

d-wojciechowski:
Hello, I would like to propose a change that will allow the user to decide if from a single field multiple validation errors may arise.
This was already reported as issue here: https://github.com/go-playground/validator/issues/1268 
## Fixes Or Enhances

### DIFF
--- a/options.go
+++ b/options.go
@@ -24,3 +24,13 @@ func WithPrivateFieldValidation() Option {
 		v.privateFieldValidation = true
 	}
 }
+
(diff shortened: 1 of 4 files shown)


fatkharrofiqi:
Hi @d-wojciechowski, just wondering why this PR was closed. I think this feature is super useful! Is there another way to achieve the same thing, or are there any plans to add it in the future? Thanks a lot!

Issue #1291, opened 2024-07-19
security: update mimetype dependency to new secure version


9ssi7:
- [x] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [x] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Issue, Question or Enhancement:

A critical security update has been released for the golang.org/x/net package.

Here are the details: https://github.com/advisories/GHSA-4v7x-pqxf-cx7m

The important part here is that the problem persists in versions 0.23.0 and below.

This problem persists with the mimetype package v1.4.3, and this problem has been resolved in v1.4.4. Check release notes for details https://github.com/gabriel-vasile/mimetype/releases/tag/v1.4.4.

What to do:

```
go get go get github.com/gabriel-vasile/mimetype@v1.4.4
```

<img width="503" alt="Screenshot 2024-07-19 at 10 08 12" src="https://github.com/user-attachments/assets/2856a604-03f2-420b-b072-e4cac1b37916">

Z-M-Huang:
This has become a sever issue with a 8.7/10 rating from snyk website. 
https://github.com/go-playground/validator/issues/1291

PR #1292, opened 2024-07-19
feat(#1291): upgrade mimetype dependency version for secure

9ssi7:
## Fixes Or Enhances

the mimetype dependency version upgraded.

Check for details https://github.com/go-playground/validator/issues/1291


9ssi7:
When I looked into it a bit, the mimetype repo started using bytes.Cut with version 1.4.4.

bytes.Cut came with version 1.18 of go. [Check for details](https://pkg.go.dev/bytes#Cut). Therefore, an error occurs in version 1.17.

We can remove 1.17 from the supported versions. Since 1.17 is already available for older versions, this will not be a problem.

9ssi7:
Yes, what I wrote in my previous comment is no longer valid. We used the module replace feature, which is a better way. This way, we can continue to support 1.17.

Now we just need to include this pr for security

nodivbyzero:
Already updated

PR #1307, opened 2024-08-16
`excluded_unless` bug fix

chargraves85:
## Fixes Or Enhances

The verbiage in the documentation for both `excluded_unless` and `required_unless` is very similar.  One would expect the opposite behavior for each when using multiple values of the same field that is being validated against.

For example:

This fails in all 3  cases in sample code
```go
package main

import (
	"fmt"
)


type TestStruct struct {
	Foo int
	Bar string `validate:"excluded_unless=Foo 1 Foo 2"`
}

func main() {

	validator := validator.New()

	testA := TestStruct{
		Foo: 1,
		Bar: "test",
	}
	testB := TestStruct{
		Foo: 2,
		Bar: "test",
	}
	testC := TestStruct{
		Foo: 3,
		Bar: "test",
	}

	err := validator.Struct(testA)
	if err != nil {
		fmt.Println(err.Error())
	}

	err = validator.Struct(testB)
	if err != nil {
		fmt.Println(err.Error())
	}

	err = validator.Struct(testC)
	if err != nil {
		fmt.Println(err.Error())
	}

}
```

However, the equivilant opposite of this only fails in TestC

```go
type TestStruct struct {
	Foo int
	Bar string `validate:"required_unless=Foo 1 Foo 2"`
}

func main() {

	validator := validator.New()

	testA := TestStruct{
		Foo: 1,
	}
	testB := TestStruct{
		Foo: 2,
	}
	
	err := validator.Struct(testA)
	if err != nil {
		fmt.Println(err.Error())
	}

	err = validator.Struct(testB)
	if err != nil {
		fmt.Println(err.Error())
	}

}
```


khoirulamri:
any update for this issue?

chargraves85:
> any update for this issue?

Looks like we have a conflict. I can update the PR but had no momentum on it when I first posted.

nodivbyzero:
Please resolve the conflicts, and I’ll proceed with the approval

chargraves85:
@nodivbyzero thank you. We should be ready.

nodivbyzero (approved):

FJJLeon:
@chargraves85 @nodivbyzero Hi there, I’m a bit confused about this bug fix and was hoping you could help clarify.  

Per the [official documentation](https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Excluded_Unless):  
> The field under validation must not be present or empty unless all the other specified fields are equal to the value following the specified field. For strings ensures value is not "". For slices, maps, pointers, interfaces, channels and functions ensures the value is not nil. For structs ensures value is not the zero value.  

Put simply: **all** of the fields specified after `excluded_unless` must match their corresponding values (i.e., a logical AND operation).  

However, this bug fix appears to have changed this behavior to require only **any** of the specified fields to match their values (i.e., a logical OR operation) instead.

PR #1335, opened 2024-11-26
handle pq.StringArray with isUnique

lukasz0707:
## Fixes Or Enhances


deankarn:
Everything @zemzale stated plus this is already supported today registering a custom type function, here is an example using the std libraries SQL types https://github.com/go-playground/validator/blob/master/_examples/custom/main.go#L26

PR #1341, opened 2024-12-06
fill gaps in test conditions

briskt:
## Fixes Or Enhances

Added some corner cases to tests to improve test coverage, but also to gain understanding of the code.


briskt:
closing because it doesn't provide any value but makes the code longer

Issue #1191, opened 2023-11-09
Alpha Space


shamir92:
- [true ] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [ true] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Issue, Question or Enhancement:
Alpha validator cannot have space. I think we need it. 

### Code sample, to showcase or reproduce:

```go

	hasPass := regexp.MustCompile(`^[a-zA-Z ]+$`).MatchString(reqData.Name)

```

trungdlp-wolffun:
I custom validator @shamir92 


```go
var regexAlphaNumSpace = regexp.MustCompile("^[ \\p{L}\\p{N}]+$")

var Ins   *validator.Validate

func init() {
  Ins = validator.New()
  Ins.RegisterValidation("alphanumspace", ValidateAlphaNumSpace)
}

func ValidateAlphaNumSpace(fl validator.FieldLevel) bool {
  return regexAlphaNumSpace.MatchString(fl.Field().String())
}
```

Use:


```go
type Request struct {
  Name string `json:"name" validate:"required,alphanumspace"`
}
```

zemzale:
Changing the current validator would be a breaking change. 

But I can see benefit for such a validator so feel free to open a PR for that

nic6200:
Hi @zemzale. Can I add a new validation for alpha spaces?

krobitaille-stash:
Would love this but I think the regex should be modified so only whitespace isn't valid?

Issue #1239, opened 2024-03-03
Regex support


Vyacheslav1557:
v10

### Enhancement:

Why dont you add a new tag for regexp, like so

```go
type Country struct {
	Name   string `validate:"required,regexp=^.{1,100}$"`
	Alpha2 string `validate:"required,regexp=^[a-zA-Z]{2}$"`
	Alpha3 string `validate:"required,regexp=^[a-zA-Z]{3}$"`
	Region string `regexp:"^Europe$|^Africa$|^Americas$|^Oceania$|^Asia$"`
}
```

I know that it won’t be possible to add support for regexp to the `validate` tag. I understand that maybe this contradicts the design, but regexp is also a basic tool that everyone needs.

deefdragon:
One of the unfortunate things here is that this will only allow top level validation using regex. 

One of the more useful regex use-cases that I can think of is validating a map's keys and or values to match a specific regex pattern.

That said, I personally disagree with the lack of regex validators in general and think there should at-least be _some way_ to use regex. I suspect an un-reasonably large chunk of users of this library have added their own regex validator implementations (myself included).

deankarn:
Hey all, it seems like that this has come back around and more people are requesting regex support again and so feel the need to comment about why it doesn't exists in the base package and also why it likely continue to remain as-is.

1. Many supported characters for regexes conflict with the parsing tag internals of this package, off the top of my head comma(,) and pipe(|) which already require special actions in validation params. With regexes only more case like this will crop up and outs handcuffs on what additions and changes can be added to this package without inadvertently making 
a breaking change.

2. Having A `regex` tag hurts usage of other features down the line such as translations and reporting back more than just the validation failure to users. For example you could have have the message `Validation failed for Name on the regex tag`(not very useful in knowing why) or with registering a specific validation, backed by a regex still, `Validation failed for Name on min_name_length tag` as an example.

3. Adding a `regex` tag would require adding some sort of mutex guarding and compiling which can cause a small amount of contention but also the possibility that a regex fails to compile at runtime when neither is strictly necessary to use regexes through registration of other tags.

4. They hurt the readability of code, some can get very large (thousands of characters long).

5. If you use the same regexes a bunch you end up copying it all over the place instead of registering a validation backed by a regex which can be reused and live, be maintained, optimized and corrected if an error/issue is found all in one place.

6. If you really want a `regex` validation tag there's nothing stopping you from adding one pretty easily.


[...]

PR #1343, opened 2024-12-07
Add alphaspace validator

takaaa220:
## Fixes Or Enhances

Add alphaspace validator for https://github.com/go-playground/validator/issues/1191.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,7 @@ validate := validator.New(validator.WithRequiredStructEnabled())
 | Tag | Description |
 | - | - |
 | alpha | Alpha Only |
+| alphaspace | Alpha Space |
 | alphanum | Alphanumeric |
 | alphanumunicode | Alphanumeric Unicode |
 | alphaunicode | Alpha Unicode |
--- a/doc.go
+++ b/doc.go
@@ -762,6 +762,12 @@ This validates that a string value contains ASCII alpha characters only
 
 	Usage: alpha
 
+# Alpha Space
+
+This validate that a string value contains ASCII alpha characters and spaces only
+
+	Usage: alphaspace
+
 # Alphanumeric
 
 This validates that a string value contains ASCII alphanumeric characters only
--- a/regexes.go
+++ b/regexes.go
@@ -7,6 +7,7 @@ import (
 
 const (
 	alphaRegexString                 = "^[a-zA-Z]+$"
+	alphaSpaceRegexString            = "^[a-zA-Z ]+$"
 	alphaNumericRegexString          = "^[a-zA-Z0-9]+$"
 	alphaUnicodeRegexString          = "^[\\p{L}]+$"
 	alphaUnicodeNumericRegexString   = "^[\\p{L}\\p{N}]+$"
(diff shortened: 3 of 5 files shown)


regexes.go
```diff
@@ -7,6 +7,7 @@ import (
 
 const (
 	alphaRegexString                 = "^[a-zA-Z]+$"
+	alphaSpaceRegexString            = "^[a-zA-Z ]+$"
```
takaaa220, line 10: Would it be better for the regex to not match if the input contains only spaces?

doc.go
```diff
@@ -762,6 +762,12 @@ This validates that a string value contains ASCII alpha characters only
 
 	Usage: alpha
 
+# Alpha Space
+
+This validate that a string value contains ASCII alpha characters and spaces only
```
ccoVeille, line 767:
```suggestion
This validates that a string value contains ASCII alpha characters and spaces only
```

regexes.go
ccoVeille: I would leave it like that

But people already asked for what you suggest, so I don't know

https://github.com/go-playground/validator/issues/1191#issuecomment-2072078625

But then, should leading and trailing space are valid?

- ` foo`, trailing space, valid or invalid?
- `foo `, leading space, valid or invalid
- `` empty, invalid?
- ` ` only space, invalid?
- `foo bar`

I'm asking because the regexp will need to be adapted

README.md
```diff
@@ -134,6 +134,7 @@ validate := validator.New(validator.WithRequiredStructEnabled())
 | Tag | Description |
 | - | - |
 | alpha | Alpha Only |
+| alphaspace | Alpha Space |
```
ccoVeille, line 137: I'm a bit annoyed, this could lead to obesity

I mean, someone will then ask for alphanumspace, alphaunicodespace, and so on

The regexps are computed by the lib even if no one use it, so lib becomes slower and slower.

The PR addresses the issue raised. So it's OK.

But I would like to step out a bit.

The regexp rule idea was already declined, for good reasons

- https://github.com/go-playground/validator/issues/1239

So I think the way to consider doing this is via the custom validator

https://github.com/go-playground/validator/issues/1191#issuecomment-1886729222

takaaa220: @ccoVeille 
Thank you for your detailed feedback. I understand and agree your concerns regarding the potential growth and its impact on library performance.

If you feel a custom validator is the better path for this functionality, I’m happy to close this pull request if necessary.

doc.go
takaaa220: fixed 
[da6de87](https://github.com/go-playground/validator/pull/1343/commits/da6de8794ce61102912028dd1df34efedf6e8c7f)

regexes.go
takaaa220: Actually, I reviewed https://github.com/go-playground/validator/issues/1191#issuecomment-2072078625 before asking the question. My question was more for clarification to ensure we’re aligned on the intended behavior. 
I think that making the regex more complex to handle edge cases like these is unnecessary. It avoids unnecessary complexity while meeting the expected use cases.

README.md
ccoVeille: I'm a random Gopher,user of the lib

Let's wait for a maintainer's feedback

takaaa220: > I'm a random Gopher,user of the lib

oh, I mistakenly thought you were the maintainer.

zemzale: Well, the regex's are lazy loaded, so if anyone doesn't use them, it doesn't affect them as much. 
 
 In this case if people think this is a good addition, I don't think we should block it because we worry about it not being used and slightly bigger binary.

zemzale (approved):
LGTM

README.md
ccoVeille: Thanks for  your explanation

rajdip-b:
I would really love to see this PR getting implemented :/ Sucks to always write my own validator for this trivial taks!

nodivbyzero:
@zemzale Would you mind checking if this PR is good to merge?

Issue #763, opened 2021-05-02
validating optional field is between certain range


thegalang:
### Issue, Question or Enhancement:

Is there a way to validate optional fields? let's say I want to validate if field `str` has either length 0 (not included) or length greater than 5.

### Code sample, to showcase or reproduce:

Currently, I made a custom validator and tag like so:
```
type TestStruct {
TestStr string `validate:"optional_gte=5"`
}
```
Is there any way to do this built-in? if not, I can submit a PR with my custom tags if it's beneficial.

BrianLeishman:
You can use "omitempty," at the beginning of your validation.

```go
type TestStruct {
TestStr string `validate:"omitempty,gte=5"`
}
```

mattwelke:
I ran into this use case today. I had a string property that was optional but if provided, must be a string with length greater than or equal to 1.

I used a string pointer for its type and `validate:"omitempty,min=1"`.

PR #1346, opened 2024-12-17
chore: using errors.As instead of type assertion

fatelei:
## Fixes Or Enhances

fix issue  #763 

### DIFF
--- a/_examples/simple/main.go
+++ b/_examples/simple/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/go-playground/validator/v10"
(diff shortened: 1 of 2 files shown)


nodivbyzero:
How does this change fix the issue #763? 
image

_examples/simple/main.go
```diff
@@ -61,24 +62,26 @@ func validateStruct() {
 		// this check is only needed when your code could produce
 		// an invalid value for validation such as interface with nil
 		// value most including myself do not usually have code like this.
-		if _, ok := err.(*validator.InvalidValidationError); ok {
+		if errors.Is(err, &validator.InvalidValidationError{}) {
```
ccoVeille, line 65: I don't think it works

This would be better

```suggestion
		if errors.As(err, &validator.InvalidValidationError{}) {
```

errors.Is is used for sentinel errors.
errors.As is used to check an error is from a certain type, and converts it

_examples/struct-level/main.go
```diff
@@ -114,33 +115,36 @@ func main() {
 		// this check is only needed when your code could produce
 		// an invalid value for validation such as interface with nil
 		// value most including myself do not usually have code like this.
-		if _, ok := err.(*validator.InvalidValidationError); ok {
+		if errors.Is(err, &validator.InvalidValidationError{}) {
```
ccoVeille, line 118: Same here

ccoVeille (changes requested):

ccoVeille (approved):

deankarn (approved):
LGTM now that it's using `errors.As`

_examples/simple/main.go
antonsoroko: unfortunately, this introduced "panic: errors: *target must be interface or implement error", because double pointer shopuld be used in this case for for `As()`.
fix in https://github.com/go-playground/validator/pull/1378

i guess this suggestion was accepted by PR author without actual run of code.

PR #1356, opened 2025-01-02
bump golang.org/x/net to v0.33.0 to fix CVE-2024-45338

sonasingh46:
Bumps golang.org/x/net to v0.33.0 to fix CVE-2024-45338. 
It also bumps mimetype lib to latest v1.4.7.

**Notes To Reviewers:**
This bump meant to also bump the go version to at least 1.20 because of the usage `subtle.XORBytes` in the crypto lib after the bump. 

This means we can no longer have the test matrix for go version lower than 1.20 otherwise it will fail with the following error: 
```
# golang.org/x/crypto/sha3
Error: ../../../go/pkg/mod/golang.org/x/crypto@v0.31.0/sha3/sha3.go:123:15: undefined: subtle.XORBytes
Error: ../../../go/pkg/mod/golang.org/x/crypto@v0.31.0/sha3/shake.go:129:22: undefined: bytes.Clone
```

See this for more reference, https://github.com/golang/go/issues/68035

**More Context:** 
The CVE is coming from the indirect dependency of golang.org/x/net from https://github.com/gabriel-vasile/mimetype. I think it is worth to update the dependency to fix the CVE and not wait for the [release](https://github.com/gabriel-vasile/mimetype/pull/620) from mimetype!


sonasingh46:
Hi @deankarn, ( Sorry for a direct tag ) -- 
Given that the current latest version of this lib has a exploitable vulnerability, I raised this bump PR to fix it. I was thinking it will be nice to get a release cut for this. I am willing to help and spend time for the release if needed.

nodivbyzero:
@sonasingh46 
Thank you very much for your contribution! 
I’ve created a similar [PR](https://github.com/go-playground/validator/pull/1339) here, but there’s an ongoing discussion regarding the minimum supported Go version.
https://github.com/go-playground/validator/discussions/1342

nodivbyzero:
Should we close this PR? 
The `golang.org/x/net` was updated to v0.34.0 in commit 2cce309b681d803db45519afc303a5d1598d3de1.

sonasingh46:
Yes. Thanks

PR #1372, opened 2025-02-12
Added support for validating OCI tags

MadsRC:
## Fixes Or Enhances

Added a new `oci_tag` to validate an OCI manifest tag as described in the OCI Distribution Spec: https://github.com/opencontainers/distribution-spec/blob/main/spec.md?plain=1#L164-L166

This is also more commonly referred to as a container tag or a docker tag.

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -241,6 +241,7 @@ var (
 		"mongodb_connection_string":     isMongoDBConnectionString,
 		"cron":                          isCron,
 		"spicedb":                       isSpiceDB,
+		"oci_tag":                       isOciTag,
 	}
 )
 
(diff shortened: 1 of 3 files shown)


nodivbyzero:
Thank you for your contribution! 
Adding a new validation function requires updating the documentation and README. 
Otherwise, everything looks good to me.

MadsRC:
I've updated my branch with latest changes from upstream and added to the README.md and doc.go file.

README.md
```diff
@@ -211,6 +215,7 @@ validate := validator.New(validator.WithRequiredStructEnabled())
 | uuid_rfc4122 | Universally Unique Identifier UUID RFC4122 |
 | md4 | MD4 hash |
 | md5 | MD5 hash |
+| oci_tag | Open Container Initiative (OCI) Tag |
```
zemzale, line 218: Can we maybe leave this and remove all the other formatting changes?

Otherwise, it doesn't feel clear and should probably be cleaned up in a different PR. There are small things like some lost/changed formatting here and there that I would rather not make part of this PR.

README.md
```diff
@@ -31,11 +32,11 @@ Installation
 
 Use go get.
 
-	go get github.com/go-playground/validator/v10
```
tjarbo: The indentation was on purpose to render the command as code. Not it is displayed as standard text.

Issue #1371, opened 2025-02-12
Feature Request: Support for Early Exit on First Validation Failure


onepiece-dz:
Description:

Hi, I'm using the go-playground/validator package and wanted to suggest a feature that could improve performance, especially when validating large structures with many fields.

Problem:
Currently, when we call validator.Validate.Struct, it performs validation on all fields, even if a field fails validation early. This can cause unnecessary overhead, especially when the structure has many fields. For example, if the first field fails validation, the subsequent fields are still validated, which can be wasteful in terms of both performance and processing time.

Feature Request:
I would like to request an enhancement that allows the validator to stop the validation process immediately when the first validation error occurs. This would provide the ability to exit early, which could significantly improve performance, especially in larger structures or cases where the first failure is the most critical.

Expected Behavior:
If a validation error occurs for any field, the validation process should terminate, and the first error should be returned.
No further fields should be validated once the first error is encountered.

Example Scenario:
type User struct {
    FirstName string `validate:"required"`
    LastName  string `validate:"required"`
    Email     string `validate:"required,email"`
}

user := &User{
    FirstName: "", // This will fail
    LastName:  "Doe",
    Email:     "john.doe@example.com",
}

err := validate.Struct(user)
// On the first validation failure (FirstName), it should stop validation
// and return the error for "FirstName".

Why this is useful:
Performance: Early exit on the first failure would save time by avoiding unnecessary checks on the remaining fields.
Flexibility: It provides the ability to stop processing on the first failure, which is a common use case for many validation scenarios (e.g., form validation, input sanitation).
Compatibility: This feature could be an opt-in behavior, allowing users to opt into "early exit" validation, without breaking current workflows.
Possible Implementation:
This could be achieved by introducing a flag or option, such as WithEarlyExit or similar, that modifies the behavior of the Struct method. When set, it would ensure that the validation process stops after the first error.

deankarn:
Hey @onepiece-dz I'm interested in learning more about your use case.

My interest lies mostly in one of your statements `It provides the ability to stop processing on the first failure, which is a common use case for many validation scenarios (e.g., form validation, input sanitation)`

One of the big reasons the library does all validation for all fields was because form validation & input sanitization was the main design driver. 

Taking From Validation as the example when filling out a form returning all the fields' issues and reporting them to the user at once to correct before attempting to submit incurring another round trip network request vs returning only one, then having to submit and make another round trip network request and having to repeat that the number of incorrect fields.

[...]

PR #1374, opened 2025-02-13
feat: add option early return in case of error

thenicolau:
## Fixes Or Enhances


thenicolau:
[Issue - 1371](https://github.com/go-playground/validator/issues/1371)

deankarn:
Asking about the use case here to learn if it's worth saving a few nanoseconds :) https://github.com/go-playground/validator/issues/1371#issuecomment-2660989420

thenicolau:
I noticed that this old PR had a merge problem, and the issue linked to it was also updated, so I fixed the issues and it is now ready to be reviewed again

zemzale (approved):
Just as mentioned before I am not sure about the use case. Not that it's invalid, but more if we actually want to add this to the library. 

But the code looks good, so approving the code change

PR #1377, opened 2025-02-17
feat: add validation for weekends

putrafajarh:
## Fixes Or Enhances
Added new validation rules `weekend`

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -220,6 +220,7 @@ var (
 		"uppercase":                     isUppercase,
 		"datetime":                      isDatetime,
 		"timezone":                      isTimeZone,
+		"weekend":                       isWeekend,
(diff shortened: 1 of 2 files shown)


putrafajarh:
Just want to know, if my PR approved, is it better to add another rule `weekday` ?
@deankarn @nodivbyzero

nodivbyzero:
Thanks for your contribution! 

Interesting idea, but the implementation seems limited to a specific date format. 
Would it still work if the date format is 'January 25, 2025 15:04' or 'Jan 25, 2025'?

putrafajarh:
@nodivbyzero it utilize time.Parse, it should work. 

The problem is according to doc, need to put 0x2C to represent comma

PR #1388, opened 2025-02-25
fixed filepath tag when used * in path

DevKhaverko:
## Fixes Or Enhances

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -1671,6 +1671,11 @@ func isFilePath(fl FieldLevel) bool {
 					// It's definitely an invalid character in the filepath.
 					return false
 				}
+
+				if t.Err == syscall.ENOENT {
+					return false
+				}
+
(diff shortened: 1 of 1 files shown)


zemzale:
Thanks for the PR, but please create an issue next time. It's impossible to understand what is the issue here.

There are few things:
1. having a `*` in a filepath is not invalid
2. `syscall.ENOENT` would indicate that the file doesn't exist, but the validator you are "fixing" doesn't support that:
> File Path [¶](https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-File_Path)
>This validates that a string value contains a valid file path but does not validate the existence of that file. This is done using os.Stat, which is a platform independent function.
3. Missing test case to support and help understand what kind of bug there even is

if you wish to check if a file exists you should use `file` validator. 

Please open an issue if you think there is an actual bug here, but without it, I will close this PR since there is not enough information.

PR #1400, opened 2025-03-19
Fix reference to parameter name in docs

yegvla:
The docs still mentioned the old name of the parameter, which threw me a bit off, this PR just updates the comment.


zemzale (approved):
Thanks for the small fix

PR #1402, opened 2025-03-22
Supply chain hardening

MadsRC:
## Fixes Or Enhances

This PR pins the versions of the various GitHub Actions used throughout this project.

### Why pin?

Version pinning is an essential security control to defend against supply chain attacks. Recently, a well-known GitHub Action (tj-actions/changed-files) was compromised to leak secrets. Such an attack would be not be successfull if downstream users had pinned their dependencies. More info can be defined here: https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised

Additional information about pinning GitHub Actions can be found here: https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide

### Updates

Dependabot is compatible with pinning GitHub Actions and will submit back PR's with new versions pinned. Additionally it should add a comment to the end of the hash (like I've done in the code) so show exactly what version is being used.

**NOTE**:  This PR includes changes from PR #1372, meaning that one should probably be merged first.


nodivbyzero:
Thanks for the effort on this! 
Please try to keep PRs focused on a single feature or change. 
Smaller, more focused PRs are easier to review and get merged faster. 

Let’s break this down if possible.

PR #1416, opened 2025-04-16
feat: implement ValidatorValuer interface feature

thommeo:
## Problem

In order to have custom type validations for generic types, every possible type has to be registered with `RegisterCustomTypeFunc` to return the underlying type. It is especially relevant for nullable types e.g. `sql.Null[T]`.

Discussed here: https://github.com/go-playground/validator/discussions/1232

## Fixes Or Enhances

* Adds a mechanism for validating generic types without using RegisterCustomTypeFunc by introducing `ValidatorValuer` interface.

Example:

```
type Nullable[T any] struct {
	Data T
}

func (t Nullable[T]) ValidatorValue() any {
	return t.Data
}

type Config struct {
	Name string `validate:"required"`
}

type Record struct {
	Config Nullable[Config] `validate:"required"`
}

func TestValidatorValuerInterface2(t *testing.T) {
	s := Record{
		Config: Nullable[Config]{},
	}
	v := New()
	errs := v.Struct(s)
	t.Fatalf("Error: %v", errs.Error())
}

// Key: 'Record.Config.Name' Error:Field validation for 'Name' failed on the 'required' tag
// Notice: no Data in the Record.Config.Name
```

### DIFF
--- a/util.go
+++ b/util.go
@@ -9,6 +9,13 @@ import (
 	"time"
 )
 
+// Valuer is an interface that allows you to expose a method on a type
+// (including generic types) that returns a value that is supposed to be validated.
+type Valuer interface {
+	// ValidatorValue returns the value that is supposed to be validated.
+	ValidatorValue() any
+}
+
 // extractTypeInternal gets the actual underlying type of field value.
 // It will dive into pointers, customTypes and return you the
 // underlying value and it's kind.
@@ -23,6 +30,13 @@ BEGIN:
 			return current, reflect.Ptr, nullable
 		}
 
+		if current.CanInterface() {
+			if v, ok := current.Interface().(Valuer); ok {
+				current = reflect.ValueOf(v.ValidatorValue())
+				goto BEGIN
+			}
+		}
+
 		current = current.Elem()
 		goto BEGIN
 
@@ -34,6 +48,13 @@ BEGIN:
 			return current, reflect.Interface, nullable
 		}
 
+		if current.CanInterface() {
+			if v, ok := current.Interface().(Valuer); ok {
+				current = reflect.ValueOf(v.ValidatorValue())
+				goto BEGIN
+			}
+		}
+
 		current = current.Elem()
 		goto BEGIN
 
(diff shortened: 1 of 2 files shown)


util.go
```diff
@@ -9,6 +9,13 @@ import (
 	"time"
 )
 
+// ValidatorValuer is an interface that allows you to expose a method on a type
+// (including generic types) that returns a value that is supposed to be validated.
+type ValidatorValuer interface {
```
zemzale, line 14: I think we can drop the `Validator`, Outside this functions it's `validator.ValidatorValuer` which stutters.

`validator.Valuer` should be fine

zemzale (dismissed):
The good looks good.

My only concerns would maybe be the naming of the interface and it's function.

vinijabes:
Do we have any updates on this?

thommeo:
> Do we have any updates on this?

Still waiting for review by @go-playground/validator-maintainers

zemzale (approved):
The code itself is good.

Just not fully sure about the whole thing just as written in the discussion, that this is very implicit API that would tie in the validator with your types. 

Would want more opinions on this before merging.

kayla-barenz-kr:
Very much looking forward to this feature. @zemzale  Any updates here?

mrWumpus:
Is this issue just waiting for more input before merging? I think this is a fine implementation for the problem. I've had to work around this so far and this simple fix would help a lot actually.

Seems the concern is that using an interface on your type would make you have to be aware of the validation framework for that type, but I'd argue that in these custom cases you have to do that any way. Especially without this fix you have to put in extra validator code for that type. Granted, it doesn't have to be directly tied to that type. But if that's an issue, you don't have to use it.

So I vote to push this through. Thanks for the fix @thommeo  and for the time to review @zemzale

nodivbyzero:
Sorry for joining the conversation late. 
I think this is a valuable feature, but the PR is missing some key pieces - specifically documentation and examples. Let's add those, and then we can move forward with merging it.

kayla-barenz-kr:
@thommeo are you able to add the documentation? or if they are unable to, is this something maintainers can add in a separate PR @nodivbyzero ?

nodivbyzero:
@kayla-barenz-kr Yep, we can add it in a separate PR 👍 
Feel free to open a new issue so we can track the request.

Issue #1413, opened 2025-04-10
Is there a way not to include name of Embedded structure in Namespace for error?


arxeiss:
We are using Validator to validate incoming JSON requests. And because all requests contains common attributes we are using embedding (`Human` in the example below).

Since that embedding has no tags (in Adult), it works exactly as we want. When `json.Unmarshal` (or json.Marshal)  is called it works as we want. In example below, `first_name` and `last_name` are in JSON as top-level attributes.

But the issue happen during validation. Because the error contains those errors:

- `"Adult.id_card_number": "Failed on tag: required"`
- `"Adult.Human.first_name": "Failed on tag: min",`
- `"Adult.address.city": "Failed on tag: required",`

The fact, that `"Adult"` is there we don't care much. It is easy just to remove everything before first `.`. 

After we do trimming, we will have `id_card_number` and `address.city` keys which are correct. They respect same structure as input JSON. But `Human.first_name` is the issue. Because in input JSON there is no `"Human"` key at all.

My question is, if there is a way not to include name of embedded structure into namespace? 

```go
package main

import (
	"bytes"
	"encoding/json"
	"fmt"
	"reflect"
	"strings"

	"github.com/go-playground/validator/v10"
)

type Human struct {
	FirstName string `json:"first_name" validate:"required,min=2"`
	LastName  string `json:"last_name" validate:"required,min=2"`
}

type Address struct {
	Street string `json:"street" validate:"required,min=2"`
	City   string `json:"city" validate:"required,min=2"`
}

// User contains user information
type Adult struct {
	Human
	Address      Address `json:"address"`
	IdCardNumber string  `json:"id_card_number" validate:"required,min=10"`
}

// use a single instance of Validate, it caches struct info
var validate *validator.Validate

func main() {
	validate = validator.New(func(v *validator.Validate) {
		v.RegisterTagNameFunc(func(fld reflect.StructField) string {
			name := strings.SplitN(fld.Tag.Get("json"), ",", 2)[0]
			// skip if tag key says it should be ignored
			if name == "-" {
				return ""
			}
			return name
		})
	})

	data := Adult{}
	err := json.Unmarshal([]byte(`{
		"first_name":"I",
		"last_name":"O",
		"address":{
			"street":"X"
		}
	}`), &data)

	if err != nil {
		panic(err)
	}

	err = validate.Struct(data)
	jsonFieldErrResponse := map[string]string{}
	jsonNsErrResponse := map[string]string{}
	if err != nil {
[...]

PR #1426, opened 2025-05-07
Feat: add tag to skip namespace in embedded structures

arxeiss:
## Enhances

Fixing #1413 and #776

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -53,6 +53,7 @@ var (
 		keysTag:           {},
 		endKeysTag:        {},
 		structOnlyTag:     {},
+		skipNamespaceTag:  {},
 		omitzero:          {},
 		omitempty:         {},
 		omitnil:           {},
--- a/cache.go
+++ b/cache.go
@@ -22,6 +22,7 @@ const (
 	typeEndKeys
 	typeOmitNil
 	typeOmitZero
+	typeSkipNamespace
 )
 
 const (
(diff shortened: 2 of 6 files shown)


ns-vchen:
@arxeiss thanks for having this PR, I'm expecting the merge
BTW, I notice that you also need to change the code [here](https://github.com/go-playground/validator/blob/fe37a2fee8befe9c7bf36c977a242a89667c2855/struct_level.go#L116).

it is widely used by the struct-level validation

deankarn:
I'm going to have to think about this one because tags are for validation or guiding validation but this adds in a new paradigm where tags outside of validations themselves can dynamically affect other aspects of the validation runtime.

There's also the matter to think about order and ability to override or needing to become a core part of the runtime like `omitempty`.

I'm thinking this maybe could be better handled in the `RegisterTagNameFunc` function. With a few minor tweaks to:
1. Allow blank values in the cache extraction.
2. Add a few checks to ensure if the alt name is blank not to append to the namespace.

Here is a quick example where I did just that custom, just for the below case a few more checks would be necessary, but you get the idea. There would also be a question of backward compatibility, technically `RegisterTagNameFunc` was already supposed to act like this but hasn't been for a long time.

image

abemedia:
@deankarn I implemented this behaviour in https://github.com/go-playground/validator/pull/1567

> I'm going to have to think about this one because tags are for validation or guiding validation but this adds in a new paradigm where tags outside of validations themselves can dynamically affect other aspects of the validation runtime.
> 
> There's also the matter to think about order and ability to override or needing to become a core part of the runtime like `omitempty`.
> 
> I'm thinking this maybe could be better handled in the `RegisterTagNameFunc` function. With a few minor tweaks to:
> 
> 1. Allow blank values in the cache extraction.
> 2. Add a few checks to ensure if the alt name is blank not to append to the namespace.
> 
> Here is a quick example where I did just that custom, just for the below case a few more checks would be necessary, but you get the idea. There would also be a question of backward compatibility, technically `RegisterTagNameFunc` was already supposed to act like this but hasn't been for a long time.
> 
> image

Issue #1424, opened 2025-05-04
[Bug]: omitzero does not work with Slice


AnalogJ:
### What happened?

omitzero should correctly ignore empty slices, instead validation continues. 

~~Looks related to this switch case for hasZero - https://github.com/go-playground/validator/blob/master/baked_in.go#L1796
It incorrectly treats `nil` as the zero case for slices.~~
Forked the code to experiment with that case, couldn't fix the issue. Looks like `omitempty` is special in some way. 

### Version

v10.26.0

#### omitempty
```go


type MyType struct {
	//OrgId   string   `validate:"required"`
	MySlice []string `validate:"omitempty,min=1,dive,uri"`
}

func (req *MyType) Validate() error {
	validate := validator.New(validator.WithRequiredStructEnabled())
	return validate.Struct(req)
}

func main() {
	myInstance1 := &MyType{
		MySlice: nil,
	}
	log.Printf("isValid: %v", myInstance1.Validate())
    // isValid: <nil>

	myInstance2 := &MyType{
		MySlice: []string{},
	}
	log.Printf("isValid: %v", myInstance2.Validate())
    // isValid: Key: 'MyType.MySlice' Error:Field validation for 'MySlice' failed on the 'min' tag
}
```

#### omitzero -- which should handle empty slice correctly

```go
type MyType struct {
	//OrgId   string   `validate:"required"`
	MySlice []string `validate:"omitzero,min=1,dive,uri"`
}

func (req *MyType) Validate() error {
	validate := validator.New(validator.WithRequiredStructEnabled())
	return validate.Struct(req)
}

func main() {
	myInstance1 := &MyType{
		MySlice: nil,
	}
	log.Printf("isValid: %v", myInstance1.Validate())
    //isValid: <nil>

	myInstance2 := &MyType{
		MySlice: []string{},
	}
	log.Printf("isValid: %v", myInstance2.Validate())
    //isValid: Key: 'MyType.MySlice' Error:Field validation for 'MySlice' failed on the 'min' tag
}
```

junaid2005p:
I think the same issue is with maps as well.

```go
type OmitEmptyExample struct {
	Data map[string]string `validate:"omitempty,min=2"`
}

type OmitZeroExample struct {
	Data map[string]string `validate:"omitzero,min=2"`
}

func main() {
	validate := validator.New()

	test1OmitEmpty := OmitEmptyExample{Data: map[string]string{}}
[...]

Issue #1430, opened 2025-05-13
[Question/Bug]: Validation not working for map fields with struct values containing validation tags


hi-rai:
### What happened?

I'm facing an issue with validating a struct that contains a map field. The map's keys and values need to be validated, but the current setup isn't working as expected.

### Struct Definitions

1.  I have a struct `Outer` with a field `Inner` that is a map of type `map[string]Inner`
2. `Outer.Inner` is of type `Inner` and has a single field `Value`

```go
type Inner struct {
	Value string `validate:"max=5"`
}

type Outer struct {
	Inner map[string]Inner `validate:"dive,keys,max=5,endkeys"`
}
```

### Issue

1. **Current Behavior**: Validation does not seem to apply to the map values (`Inner` structs) as expected.
2. **Workarounds**
    a. Adding `,required` after `endkeys` seems to trigger the validation, but it also prevents ([with validator.WithRequiredStructEnabled()](https://github.com/go-playground/validator#special-notes)) empty struct values (`Inner{}`), which is not the desired behavior.
    b. Removing key validations and retaining only `dive` also triggers validation on values (but key validations are missed)
3. **Documentation**: I did not find any similar mention or examples in the documentation.

Is this a bug ? Or am I supposed to do something differently to achieve my desired behavior ?

### Version

v10.26.0

### Example Code

```go
package main

import (
	"fmt"

	"github.com/go-playground/validator/v10"
)

type Inner struct {
	Value string `validate:"max=5"`
}

type Outer1 struct {
	Inner map[string]Inner `validate:"dive,keys,max=5,endkeys"`
}

type Outer2 struct {
	Inner map[string]Inner `validate:"dive"`
}

type Outer3 struct {
	Inner map[string]Inner `validate:"dive,keys,max=5,endkeys,required"`
}

func main() {
	v := validator.New(validator.WithRequiredStructEnabled())

	// Should fail, but passes
	fmt.Println("Outer1:", v.Struct(Outer1{Inner: map[string]Inner{
		"a": {Value: "1234567890"},
	}}))

	// Should fail, and fails (but there is no validation for keys)
	fmt.Println("Outer2:", v.Struct(Outer2{Inner: map[string]Inner{
		"a": {Value: "1234567890"},
	}}))

	// Should fail, and fails (but has extra required validation, which is not intended)
	fmt.Println("Outer3:", v.Struct(Outer3{Inner: map[string]Inner{
		"a": {Value: "1234567890"},
	}}))

[...]

PR #1436, opened 2025-06-03
Omitzero does not work with slice and map bug

junaid2005p:
## Fixes 

Fixed a bug where using omitzero still validated empty structs and maps. This PR ensures that if omitzero is used as a tag then empty slices and maps are not validated.

* Modified the `baked_in.go` file to ensure that slices and maps are considered non zero only when they are not nil and they are not empty.
* Added tests to ensure empty slices and maps are not validated while using omitzero tag.

Check #1424 for more details.

### DIFF
--- a/validator_test.go
+++ b/validator_test.go
@@ -14633,3 +14633,123 @@ func TestMapEmptyPointerStructValueNoError(t *testing.T) {
 	errs = validate.Struct(obj2)
 	Equal(t, errs, nil)
 }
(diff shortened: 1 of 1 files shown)


junaid2005p:
Hi @nodivbyzero would you kindly take a look if you get some time? Thanks

junaid2005p:
Thanks @nodivbyzero for the review. Can we merge this PR if everything looks good? Also can we work on closing issues #1424 and #1430 ?

Issue #1437, opened 2025-06-05
[Bug]: validator panic with []byte and required_if


qiuchengxuan:
### What happened?

Program panic when using []byte with required_if

```
panic: strconv.ParseInt: parsing "nil": invalid syntax

goroutine 1 [running]:
github.com/go-playground/validator/v10.panicIf(...)
        /home/qiubuyi/.go/pkg/mod/github.com/go-playground/validator/v10@v10.26.0/util.go:294
github.com/go-playground/validator/v10.asInt({0x5b6e9d?, 0x5d6380?})
        /home/qiubuyi/.go/pkg/mod/github.com/go-playground/validator/v10@v10.26.0/util.go:229 +0x48
github.com/go-playground/validator/v10.requireCheckFieldValue({0x6551b0, 0xc00016a7e0}, {0x5b6e97, 0x5}, {0x5b6e9d, 0x3}, 0x0)
        /home/qiubuyi/.go/pkg/mod/github.com/go-playground/validator/v10@v10.26.0/baked_in.go:1868 +0x3b1
github.com/go-playground/validator/v10.requiredIf({0x6551b0, 0xc00016a7e0})
        /home/qiubuyi/.go/pkg/mod/github.com/go-playground/validator/v10@v10.26.0/baked_in.go:1893 +0xac
github.com/go-playground/validator/v10.New.wrapFunc.func2({0xc00016a7e0?, 0x5bedc0?}, {0x6551b0?, 0xc00016a7e0?})
        /home/qiubuyi/.go/pkg/mod/github.com/go-playground/validator/v10@v10.26.0/baked_in.go:44 +0x27
github.com/go-playground/validator/v10.(*validate).traverseField(0xc00016a7e0, {0x654030, 0x7d6ea0}, {0x5d6380?, 0xc0001cc0c0?, 0xc0000a2d50?}, {0x5bedc0?, 0xc0001cc0c0?, 0x0?}, {0xc000182280, ...}, ...)
        /home/qiubuyi/.go/pkg/mod/github.com/go-playground/validator/v10@v10.26.0/validator.go:473 +0x11e3
github.com/go-playground/validator/v10.(*validate).validateStruct(0xc00016a7e0, {0x654030, 0x7d6ea0}, {0x5b7f40?, 0xc0001cc0c0?, 0x1306a8?}, {0x5d6380?, 0xc0001cc0c0?, 0xc0001cc090?}, {0x656cd8, ...}, ...)
[...]

PR #1442, opened 2025-06-18
Fix: Validator panics when 'nil' is used along with required if for slices and maps

junaid2005p:
## Fixes 
Added support to check for 'nil' values in case of maps and slices when used with required if. (Previously only lengths were being checked as a result using nil with required if would lead to a panic since the program tried to convert 'nil' to an integer)
Created a separate case for arrays to only check lengths since arrays cannot be nil in go.

Check #1437 for more details.

### DIFF
--- a/validator_test.go
+++ b/validator_test.go
@@ -14753,22 +14753,6 @@ func TestOmitZeroWithMaps(t *testing.T) {
 		Equal(t, err2, nil) // No error
 	})
 }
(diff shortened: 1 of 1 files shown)


junaid2005p:
@nodivbyzero kindly take a look and let me know if I need to change anything else

nodivbyzero (approved):

junaid2005p:
@nodivbyzero, if everything looks good, can we merge?

PR #1468, opened 2025-09-17
fix: should panic when define duplicate field param in `required_if`

### LINKED ISSUE #1466: [Bug]: required_if with multiple values on the same field unexpected behavior
### What happened?

When using required_if with the same field and multiple comparison values, the validator unexpectedly passes, it should expect error

### Example:
```go
type Request struct {
    Type  string `validate:"required"`
    Value string `validate:"required_if=Type A Type B"`
}
```

`Type` value cannot be A and B at the same time, so the validator will pass in anycase

Minimal Repro (failing test)
```go
package demo

import (
    "testing"

    "github.com/stretchr/testify/require"
)

type Request struct {
    Type  string `validate:"required"`
    Value string `validate:"required_if=Type A Type B"`
}

func TestValidateRequiredIf(t *testing.T) {
    v := validator.New()

    // Type A, Value empty -> should fire panic error
    r := Request{Type: "A", Value: ""}
    err := v.Struct(r)
    require.Error(t, err, "expected error: Value must be required when Type is A or B")

    // Type B, Value empty -> should fire panic error
    r = Request{Type: "B", Value: ""}
    err = v.Struct(r)
    require.Error(t, err, "expected error: Value must be required when Type is A or B")
}
```

### Expected behavior
This tag `validate:"required_if=Type A Type B"` should output panic error, and only accept unique field name

### Actual behavior

Validation passes (no error) in the above cases

If this makes sense, I’d be happy to help contribute an implementation

### Version

v10.12.0

If this makes sense, I’d be happy to help contribute fix

duyquang6:
## Fixes Or Enhances
fix #1466 

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -1872,6 +1872,15 @@ func requiredIf(fl FieldLevel) bool {
 	if len(params)%2 != 0 {
 		panic(fmt.Sprintf("Bad param number for required_if %s", fl.FieldName()))
 	}
+
+	seen := make(map[string]struct{})
+	for i := 0; i < len(params); i += 2 {
+		if _, ok := seen[params[i]]; ok {
+			panic(fmt.Sprintf("Duplicate param %s for required_if %s", params[i], fl.FieldName()))
+		}
+		seen[params[i]] = struct{}{}
+	}
+
 	for i := 0; i < len(params); i += 2 {
 		if !requireCheckFieldValue(fl, params[i], params[i+1], false) {
 			return true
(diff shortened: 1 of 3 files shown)


nodivbyzero:
Looks good to me. Just a couple of follow-ups:

1. We should probably add the same logic to required_unless.
2. Please update `doc.go` to reflect the new behavior.

duyquang6:
@nodivbyzero 
- Updated doc.go
- required_unless is a little bit different, I found that it work fine with duplicate params, so I just add testcase for it

nodivbyzero (approved):

Issue #1471, opened 2025-10-07
Feature request: Support for ISO 9362:2022 BIC (SWIFT) codes


fira42073:
**Problem**
The built-in `bic` validator currently appears to reject some valid BICs introduced under the ISO 9362:2022 revision — specifically those where the first four characters of the party prefix can include digits (e.g. `E097AEXX`).

This pattern is valid per ISO 9362:2022, section 6.3.1, which allows the party prefix to be **alphanumeric**, not just letters.

**Example**
```go
type Bank struct {
    SwiftCode string `validate:"bic"`
}

bank := Bank{SwiftCode: "E097AEXX"} // Valid under ISO 9362:2022
err := validate.Struct(bank)
```
Currently, this fails validation with the built-in bic tag.

Expected behavior
The bic tag should validate codes conforming to both:
ISO 9362:2014 (legacy 8- or 11-char codes), and
ISO 9362:2022 (new alphanumeric party prefix).

References
[ISO 9362:2022 (Business Identifier Code)](https://www.iso.org/standard/84108.html)
Example real-world BIC: E097AEXX
Related discussion: [Symfony issue 51975](https://github.com/symfony/symfony/issues/51975)

I'm willing to contribute here.

fira42073:
Apparently this is the regex string
```
bicRegexString                   = `^[A-Za-z]{6}[A-Za-z0-9]{2}([A-Za-z0-9]{3})?$`
```

changing it to the following one should fix the problem
```
bicRegexString                   = `^[A-Z0-9]{4}[A-Z]{2}[A-Z0-9]{2}(?:[A-Z0-9]{3})?$`
```

But I cannot say I'm a pro at regex, so please someone double check this C:

nodivbyzero:
@fira42073 PRs are always welcome! Just open one with an updated unit test for the new ISO 9362:2022 revision.

fira42073:
[...]

PR #1478, opened 2025-10-14
Support for ISO 9362:2022 BIC (SWIFT) codes

fira42073:
## Fixes Or Enhances
https://github.com/go-playground/validator/issues/1471

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -1378,12 +1378,19 @@ More information on https://pkg.go.dev/golang.org/x/text/language
 
 	Usage: bcp47_language_tag
 
-BIC (SWIFT code)
+BIC (SWIFT code - 2022 standard)
 
-This validates that a string value is a valid Business Identifier Code (SWIFT code), defined in ISO 9362.
-More information on https://www.iso.org/standard/60390.html
+This validates that a string value is a valid Business Identifier Code (SWIFT code), defined in ISO 9362:2022.
+More information on https://www.iso.org/standard/84108.html
 
 	Usage: bic
+
+BIC (SWIFT code - 2014 standard)
+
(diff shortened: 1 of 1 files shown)


nodivbyzero:
Would it be better to introduce a new validator, such as  `bic_iso_9362`, rather than modifying the existing one? This approach would preserve backward compatibility and ensure that current functionality remains unaffected.

fira42073:
sounds good, I can do that, but fyi https://www.iso.org/standard/60390.html the standard is withdrawn, and the 2022 one is inclusive of all old BICs from 2014 one. but if we are to proceed with the namespaced versions, how should we distinguish 2014 and 2022 ones?  `bic_iso_9362_2014` and  `bic_iso_9362_2022`?

nodivbyzero:
Another option is to rename the existing one to `bic_iso_9362_2014` and keep your current changes. We can also add a note/comment in the documentation to clarify this distinction.

fira42073:
@nodivbyzero Please check at your convenience C:
Thanks!!

nodivbyzero:
@fira42073 
Looks great! Just need to update one more place, then it's ready to merge.
https://github.com/go-playground/validator/blob/6a38036e64751ad983b154546ff08df2bff9cc28/doc.go#L1381-L1386

fira42073:
Added, thanks!
@nodivbyzero

fira42073:
rebased master, but lint is failing for some reason :confused:

nodivbyzero:
> rebased master, but lint is failing for some reason 😕

Don't worry, it's not caused by your changes - I've opened a [PR](https://github.com/go-playground/validator/pull/1487) to fix it.

nodivbyzero:
Please rebase, it should be fixed now

fira42073:
@nodivbyzero thanks! please merge at your convenience then

Issue #1348, opened 2024-12-18
bug: unix_addr check is useless


aMOPel:
- [ x] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [ x] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version eg. v9, v10: 
v10.23.0 

### Issue, Question or Enhancement:

The unix_addr check is useless. I looked into the src:

https://github.com/go-playground/validator/blob/6c3307e6c64040ebc0efffe9366927c68146ffba/baked_in.go#L2564C16-L2564C31

```go
func isUnixAddrResolvable(fl FieldLevel) bool {
	_, err := net.ResolveUnixAddr("unix", fl.Field().String())

	return err == nil
}
```

and in stdlib net:

https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/net/unixsock.go;l=57

```go
func ResolveUnixAddr(network, address string) (*UnixAddr, error) {
	switch network {
	case "unix", "unixgram", "unixpacket":
		return &UnixAddr{Name: address, Net: network}, nil
	default:
		return nil, UnknownNetworkError(network)
	}
}
```

effectivly this will never throw an error since it you're always passing "unix".

If fact, ResolveUnixAddr is itself useless to check if the socket actually exists.

You need something that checks if the file exists and if that file is a socket:

```go
func IsSocket(filePath string) (err error) {
	stats, err := os.Stat(filePath)
	if err != nil {
		err = fmt.Errorf("check file existence: %w", err)
		return
	}
	if stats.Mode().Type() != fs.ModeSocket {
		err = fmt.Errorf("not socket file: %v", filePath)
		return
	}
	return
}
```

nodivbyzero:
I'm not entirely sure about the original design intention for this validator. It might have been intended to validate input values exclusively for the 'Unix' network. However, it would be beneficial to extend its functionality to also handle `unixgram` and `unixpacket` networks. 

We always welcome pull requests with improvements or fixes.

I suggest creating a new `socket` validator and implementing it with the approach you provided.

nf-brentsaner:
A "socket" does not inherently mean a UDS/IPC socket. `tcp,127.0.0.1:1234` is a "socket".

If anything, it explicitly should be called "ipc" ("Inter-Process Communication") or "uds" (UNIX Domain Socket).

`socket` as a validator name should be applicable to all of the above cases (see also #1327).

[...]

PR #1482, opened 2025-10-30
Feat: uds_exists validator

barash-asenov:
## Fixes Or Enhances

As discussed in: https://github.com/go-playground/validator/issues/1348

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -53,4 +53,3 @@ linters:
     - varnamelen
     - wrapcheck
     - wsl
-    - modernize


validator_test.go
```diff
@@ -2950,6 +2952,76 @@ func TestUnixAddrValidation(t *testing.T) {
...
+		}
+		defer func() {
+			_ = abstractListener.Close()
+		}()
+		errs = validate.Var(abstractSockName, "uds_exists")
+		Equal(t, errs, nil)
+		errs = validate.Var("@nonexistent_abstract_socket", "uds_exists")
+		NotEqual(t, errs, nil)
+		AssertError(t, errs, "", "", "", "", "uds_exists")
+	}
```
zemzale: Only issue I have here, is that I would want the tests to be split up and ran as sub tests.

```go
t.Run("non_existant", func(t *testing.T) {
	errs = validate.Var("/tmp/nonexistent.sock", "uds_exists")
	NotEqual(t, errs, nil)
	AssertError(t, errs, "", "", "", "", "uds_exists")
})
```

Otherwise with all the `defer` and other things feel hard to read.

zemzale (dismissed):
I would like the tests to be cleaner, but otherwise looks good

validator_test.go
barash-asenov: Good idea, I've adjusted the tests accordingly.

barash-asenov:
I think I should've avoided rebasing..

PR #1489, opened 2025-11-08
Add bcp47_strict_language_tag validator

### LINKED ISSUE #1221: bcp47_language_tag doesn't fail on some non-BCP47 tags

v10

When using `bcp47_language_tag` for validation, some non-BCP47 tags such as "`eng`" or "`en_US`" are passing as valid. 

`‎isBCP47LanguageTag()` uses `golang.org/x/text/language`'s [`Parse`](https://pkg.go.dev/golang.org/x/text/language#Parse) and its documentation says:

>  [snip]
> It accepts tags in the BCP 47 format **and extensions to this standard defined in https://www.unicode.org/reports/tr35/#Unicode_Language_and_Locale_Identifiers**.

I expect both of these to fail, but they don't:

```go
package main

import (
	"fmt"
)

func main() {
	validate := validator.New()

	err := validate.Var("en_US", "bcp47_language_tag")
	if err != nil {
		fmt.Println(err.Error())
		return
	}

	err = validate.Var("eng", "bcp47_language_tag")
	if err != nil {
		fmt.Println(err.Error())
		return
	}
}
```

bfabio:
Add strict BCP47 language tag validator enforcing RFC5646 and rejecting Unicode extensions unlike language.Parse().

## Fixes Or Enhances

Fix #1221.


bfabio:
@nodivbyzero rebased

nodivbyzero:
@bfabio Thanks for contributing!

Feels like `isBCP47StrictLanguageTag` still needs a bit more refactoring:
1. Move regex to global
2. Reduce ToLower/ToUpper
3. Early rejection

Something like that:
```
var languageTagRe = regexp.MustCompile(strings.Join([]string{ ... }, ""))

func isBCP47StrictLanguageTag(fl FieldLevel) bool {
	field := fl.Field()

	if field.Kind() != reflect.String {
	}

	raw := field.String()
	upper := strings.ToUpper(raw)
	lower := strings.ToLower(raw)
	lowerTagDash := lower + "-"

	m := languageTagRe.FindStringSubmatch(upper)
	if m == nil {
		return false
	}

	// reuse lower/upper everywhere instead of recomputing
	...
}
``` 

Also, is it possible to completely replace regex?
```
tag, err := language.Parse(languageTag)
```
and then validate against IANA constraints manually.

bfabio:
Thanks for the review

> Feels like `isBCP47StrictLanguageTag` still needs a bit more refactoring:
> 
>     1. Move regex to global

Done
 
>     2. Reduce ToLower/ToUpper

I refactored it a little bit more so they're not necessary. I don't even know why I did upcase it, but it was a while ago :shrug: 
 
>     3. Early rejection

Done.

> Also, is it possible to completely replace regex?
> 
> ```
> tag, err := language.Parse(languageTag)
> ```
> 
> and then validate against IANA constraints manually.

I'm not sure what you mean here, a custom parser?

nodivbyzero:
I reviewed your PR again and the changes look good overall. 
However, I found one test case where your function is failing. 
Here's the tag - `en-rozaj-biske-1994`.
It should be valid.

bfabio:
https://www.rfc-editor.org/rfc/rfc5646.html#section-3.1.8
>    The field 'Prefix' contains a valid language tag that is RECOMMENDED
   as one possible prefix to this record's subtag, perhaps with other
   subtags.  That is, when including an extended language or a variant
   subtag that has at least one 'Prefix' in a language tag, the
   resulting tag SHOULD match at least one of the subtag's 

Holy cow, you're right. Nicely spotted!

bfabio:
@nodivbyzero fixed

nodivbyzero (approved):

zemzale (approved):

Issue #1481, opened 2025-10-30
[Bug]: Numeric arrays pass string validations


luckv:
### What happened?

Array of integers pass validations for strings.

Tested with all [string tags validation](https://pkg.go.dev/github.com/go-playground/validator/v10#readme-strings).

Go playground with all strings validations tested (they all pass)
https://go.dev/play/p/cky_4b8iONw

Maybe I misinterpreted something, or this is a strange bug?

### Version

 Tested with 2 versions:
 - 10.22.0
 - 10.28.0

### Example Code

```go
package main

import (
	"log"

	"github.com/go-playground/validator/v10"
)

func main() {
	validate := validator.New(validator.WithRequiredStructEnabled())

	arr := []int{3000}

	tagsToTest := [...]string{
		"alpha",
		"alphaspace",
		"alphanum",
		"alphanumunicode",
		"alphaunicode",
		"ascii",
		"boolean",
		"contains",
		"containsany",
		"containsrune",
		"endsnotwith",
		"endswith",
		"excludes",
		"excludesall",
		"excludesrune",
		"lowercase",
		"multibyte",
		"number",
		"numeric",
		"printascii",
		"startsnotwith",
		"startswith",
		"uppercase",
	}

	for i := range tagsToTest {
		tag := tagsToTest[i]

		if err := validate.Var(arr, "printascii"); err != nil {
			log.Printf("%s: NOT PASSED - %s\n", tag, err.Error())
		} else {
			log.Printf("%s: PASSED\n", tag)
		}
	}
}
```

nodivbyzero:
@luckv It looks like your example loops through the `tagsToTest` list, but ends up testing only the `printascii` tag.
```
if err := validate.Var(arr, "printascii"); err != nil {
```

luckv:
@nodivbyzero  I'm so sorry, I forgot to pass that current tag of the loop in the validation.

Only `ascii` and  `printascii` present this bug and pass integers arrays as valid strings.
I corrected the example code: https://go.dev/play/p/ket9OEcElO_M

nodivbyzero:
@luckv Good catch! 
`isPrintableASCII` and `isASCII` do have a bug and mishandles the input as an array. 

A PR with a fix would be greatly appreciated.

luk3skyw4lker:
I'll open a PR for that in a few days 👀

PR #1498, opened 2025-11-28
Fix/1481 skip invalid type validations

KaranLathiya:
#1481 

Only strings can pass validations for strings other types return error in 'ascii' & 'printascii' validations.

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -538,7 +538,7 @@ func isPrintableASCII(fl FieldLevel) bool {
 	if field.Kind() == reflect.String {
 		return printableASCIIRegex().MatchString(field.String())
 	}
-	panic(fmt.Sprintf("Bad field type %s", field.Type()))
+	return false
 }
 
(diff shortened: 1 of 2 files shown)


baked_in.go
```diff
@@ -534,12 +534,20 @@ func hasMultiByteCharacter(fl FieldLevel) bool {
...
 }
 
 // isASCII is the validation function for validating if the field's value is a valid ASCII character.
 func isASCII(fl FieldLevel) bool {
-	return aSCIIRegex().MatchString(fl.Field().String())
+	field := fl.Field()
+	if field.Kind() == reflect.String {
+		return aSCIIRegex().MatchString(field.String())
+	}
+	panic(fmt.Sprintf("Bad field type %s", field.Type()))
```
nodivbyzero: Instead of calling `panic()`, let's return `false` so we don't break existing behavior.
```suggestion
	return false
```

PR #1507, opened 2025-12-18
fix: resolve panic when using aliases with OR operator

### LINKED ISSUE #766: Aliases or'ed with other functions causes a panic `Undefined validation function`

Custom validation aliases with "or" (`|`) panic with `Undefined validation function`

I use a custom alias for custom identifiers, that's an alias including a custom function

```go
const rawBase64URLRegexString = "^[A-Za-z0-9-_]*$"
var rawBase64URLRegex = regexp.MustCompile(rawBase64URLRegexString)
v.validate.RegisterValidation("rawbase64url", func(fl validator.FieldLevel) bool {
	return rawBase64URLRegex.MatchString(fl.Field().String())
})
v.validate.RegisterAlias("bid2", "rawbase64url,len=11")
```

```go
var req struct {
	Order string `binding:"numeric|bid2"`
}
```

This results in a panic with `Undefined validation function`. However, this does not result in a panic and works great:

```go
var req struct {
	Order string `binding:"bid2"`
}
```

Using a custom function (not an alias) also seems to work without issues

```go
var req struct {
	Order string `binding:"rawbase64url|numeric"`
}
```

Upon further testing, it seems that even built in aliases fail when or'ed. The following also panics with `Undefined validation function 'iscolor' on field 'Order' `

```go
var req struct {
	Order string `binding:"numeric|iscolor"`
}
```

shindonghwi:
Fixes #766

Using aliases with OR operator causes panic:

```go
validate.RegisterAlias("bid2", "numeric")
// validate:"numeric|bid2" → panic: Undefined validation function 'bid2'
```

The issue is that alias lookup only happens for comma-separated tags, not OR-separated ones. Fixed by adding alias check in the OR parsing logic.

Added tests for various cases (built-in aliases, custom aliases, different positions, etc).

### DIFF
--- a/cache.go
+++ b/cache.go
@@ -289,6 +289,24 @@ func (v *Validate) parseFieldTagsRecursive(tag string, fieldName string, alias s
 				if wrapper, ok := v.validations[current.tag]; ok {
 					current.fn = wrapper.fn
 					current.runValidationWhenNil = wrapper.runValidationOnNil
+				} else if aliasTag, isAlias := v.aliases[current.tag]; isAlias {
+					aliasFirst, aliasLast := v.parseFieldTagsRecursive(aliasTag, fieldName, current.tag, true)
+
+					current.tag = aliasFirst.tag
+					current.fn = aliasFirst.fn
+					current.runValidationWhenNil = aliasFirst.runValidationWhenNil
+					current.hasParam = aliasFirst.hasParam
+					current.param = aliasFirst.param
+					current.typeof = aliasFirst.typeof
+					current.hasAlias = true
+
+					if aliasFirst.next != nil {
+						nextInChain := current.next
+						current.next = aliasFirst.next
+						aliasLast.next = nextInChain
+						aliasLast.isBlockEnd = false
+						current = aliasLast
+					}
 				} else {
 					panic(strings.TrimSpace(fmt.Sprintf(undefinedValidation, current.tag, fieldName)))
 				}
--- a/validator_test.go
+++ b/validator_test.go
@@ -696,6 +696,68 @@ func TestAliasTags(t *testing.T) {
 	PanicMatches(t, func() { validate.RegisterAlias("exists!", "gt=5,lt=10") }, "Alias 'exists!' either contains restricted characters or is the same as a restricted tag needed for normal operation")
 }
 
+func TestAliasWithOrOperator(t *testing.T) {
+	validate := New()
+	validate.RegisterAlias("customnum", "numeric")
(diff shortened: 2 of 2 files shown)


cache.go
```diff
@@ -289,6 +289,34 @@ func (v *Validate) parseFieldTagsRecursive(tag string, fieldName string, alias s
...
+						nextInChain := current.next
+						// Link the rest of the alias chain
+						current.next = aliasFirst.next
+						// Connect the last alias tag to what was after current
+						aliasLast.next = nextInChain
+						// Clear isBlockEnd since this may not be the last tag in the OR expression
+						aliasLast.isBlockEnd = false
+						// Update current to point to the last tag in the alias chain
+						current = aliasLast
+					}
```
zemzale: The comments are a bit excessive.  
```suggestion
				} else if aliasTag, isAlias := v.aliases[current.tag]; isAlias {
					// Handle alias within OR expression (issue #766)
					// Recursively parse the alias tag
					aliasFirst, aliasLast := v.parseFieldTagsRecursive(aliasTag, fieldName, current.tag, true)
					
					// Note: isBlockEnd is NOT copied here; it will be set at the end of the OR loop
					current.tag = aliasFirst.tag
					current.fn = aliasFirst.fn
					current.runValidationWhenNil = aliasFirst.runValidationWhenNil
					current.hasParam = aliasFirst.hasParam
					current.param = aliasFirst.param
					current.typeof = aliasFirst.typeof
					current.hasAlias = true


					// If alias expanded to multiple tags, insert them into the chain
					if aliasFirst.next != nil {
						nextInChain := current.next
						current.next = aliasFirst.next
						aliasLast.next = nextInChain
						aliasLast.isBlockEnd = false
						current = aliasLast
					}
```

validator_test.go
```diff
@@ -696,6 +696,104 @@ func TestAliasTags(t *testing.T) {
...
+	// Test 7: Another built-in alias with OR (country_code)
+	type Test7 struct {
+		Field string `validate:"numeric|country_code"`
+	}
+
+	t7 := Test7{Field: "US"}
+	errs = validate.Struct(t7)
+	Equal(t, errs, nil) // Should pass - valid country code
+}
+
```
zemzale: Again the comments are a bit too much. None of them are needed, or should just be used as names for `t.Run` subtests

zemzale (approved):
Generally looks like a good fix, just some style issues from me.

validator_test.go
shindonghwi: Done! Removed all comments and converted tests to use t.Run subtests. @zemzale

PR #1530, opened 2026-02-10
fix: prevent panic in unique validation with nil pointer elements

### LINKED ISSUE #1319: unique tag panic on  slice contains nil elem
i want to make sure elem of pointer slice is not nil，meanwhile, i also need the Name field is unique.

the former can be achieved like: 
struct {
		F1 []*A `validate:"dive,required"`
	}
 
however, the unique tag  on nil elem cause panic

func TestName(t *testing.T) {
	type A struct {
		Name string
	}
	e := struct {
		F1 []*A `validate:"unique=Name"`
	}{F1: []*A{nil}}
	err := validate.Struct(e)
	assert.Error(t, err)
}

veeceey:
## Summary

- Fixes a panic (`reflect: call of reflect.Value.FieldByName on zero Value`) when the `unique` tag validates slices or maps containing nil pointer elements.
- Three code paths in `isUnique` are affected: (1) slices without a field param (`unique`), (2) slices with a field param (`unique=Name`), and (3) maps with pointer values. All three now safely handle nil elements.
- Nil elements are treated as equal to each other for uniqueness purposes: a single nil is valid, but multiple nils in the same collection fail the `unique` check.

Fixes #1319

## Reproduction

```go
type A struct {
    Name string
}
e := struct {
    F1 []*A `validate:"unique=Name"`
}{F1: []*A{nil}}
err := validate.Struct(e) // panics before this fix
```

## Changes

- **`baked_in.go`**: In all three branches of `isUnique` (slice no-param, slice with-param, map), check `e.IsValid()` after `reflect.Indirect` and skip nil elements with a counter. Return false when `nilCount > 1` (duplicate nils are not unique).
- **`validator_test.go`**: Added `TestUniqueValidationNilPtrSlice` with 9 sub-tests covering single nil, duplicate nils, and mixed nil/non-nil for all three code paths.

## Test plan

- [x] New test `TestUniqueValidationNilPtrSlice` passes (9 sub-tests)
- [x] Full test suite passes (`go test ./...` -- all 24 packages OK, zero failures)
- [x] No regressions in existing `TestUniqueValidation`, `TestUniqueValidationStructSlice`, and `TestUniqueValidationStructPtrSlice`

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -347,11 +347,17 @@ func isUnique(fl FieldLevel) bool {
 
 		if param == "" {
 			m := reflect.MakeMap(reflect.MapOf(elem, v.Type()))
+			var nilCount int
 
 			for i := 0; i < field.Len(); i++ {
-				m.SetMapIndex(reflect.Indirect(field.Index(i)), v)
+				e := reflect.Indirect(field.Index(i))
+				if !e.IsValid() {
+					nilCount++
+					continue
+				}
+				m.SetMapIndex(e, v)
 			}
-			return field.Len() == m.Len()
+			return field.Len()-nilCount == m.Len() && nilCount <= 1
 		}
 
 		sf, ok := elem.FieldByName(param)
@@ -366,14 +372,20 @@ func isUnique(fl FieldLevel) bool {
 
 		m := reflect.MakeMap(reflect.MapOf(sfTyp, v.Type()))
 		var fieldlen int
+		var nilCount int
 		for i := 0; i < field.Len(); i++ {
-			key := reflect.Indirect(reflect.Indirect(field.Index(i)).FieldByName(param))
+			e := reflect.Indirect(field.Index(i))
+			if !e.IsValid() {
+				nilCount++
+				continue
+			}
+			key := reflect.Indirect(e.FieldByName(param))
 			if key.IsValid() {
 				fieldlen++
 				m.SetMapIndex(key, v)
 			}
 		}
-		return fieldlen == m.Len()
+		return fieldlen == m.Len() && nilCount <= 1
 	case reflect.Map:
 		var m reflect.Value
 		if field.Type().Elem().Kind() == reflect.Ptr {
@@ -382,11 +394,17 @@ func isUnique(fl FieldLevel) bool {
 			m = reflect.MakeMap(reflect.MapOf(field.Type().Elem(), v.Type()))
 		}
 
+		var nilCount int
 		for _, k := range field.MapKeys() {
-			m.SetMapIndex(reflect.Indirect(field.MapIndex(k)), v)
+			e := reflect.Indirect(field.MapIndex(k))
+			if !e.IsValid() {
+				nilCount++
+				continue
+			}
+			m.SetMapIndex(e, v)
 		}
 
-		return field.Len() == m.Len()
+		return field.Len()-nilCount == m.Len() && nilCount <= 1
 	default:
 		if parent := fl.Parent(); parent.Kind() == reflect.Struct {
 			uniqueField := parent.FieldByName(param)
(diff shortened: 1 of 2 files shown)


nodivbyzero:
Thanks for the contribution!

I took a look at your changes - what do you think about treating `nil` as just another comparable key and dropping `nilCount` altogether? 
Something like:
```
if !e.IsValid() {
			m.SetMapIndex(zero, reflect.ValueOf(struct{}{}))
			continue
		}
```
Then we could keep the original check:
```
return field.Len() == m.Len()
```

veeceey:
Thanks for the suggestion! I considered that approach but there's a subtle correctness issue — if we use a zero-value key for nil pointers, it would collide with an actual element whose fields are all zero-valued. For example, `[]*Inner{nil, {Name: ""}}` would incorrectly fail the unique check because both nil and the zero-value struct would map to the same key.

The `nilCount` approach avoids that by keeping nil tracking completely separate from the map, so nil elements only compare against other nil elements.

nodivbyzero:
@veeceey please take a look at my PR: https://github.com/go-playground/validator/pull/1532

veeceey:
Hey @nodivbyzero, thanks for putting that together! I took a look at #1532. Your approach is definitely cleaner in terms of not needing a separate counter.

One thing I noticed though - using `reflect.Zero(elem)` as the map key for nil values means nil pointers and actual zero-value structs would collide. For example, `[]*Inner{nil, {Name: ""}}` would fail the unique check since both would map to the same zero-value key, even though they're arguably different values (nil vs an actual empty struct).

That said, I can see the argument that treating nil as equivalent to zero-value might be reasonable depending on the use case. Happy to defer to the maintainers on which behavior they prefer - just wanted to flag the semantic difference.

veeceey:
Hi maintainers, just checking in on this one. I've had a good discussion with @nodivbyzero about the approach (nilCount vs zero-value key). My approach keeps nil tracking separate to avoid collisions between nil pointers and zero-value structs. @nodivbyzero opened an alternative in #1532. Would love a maintainer's take on which semantic approach you prefer — happy to adjust either way\!

nodivbyzero:
@veeceey Is there a test case that succeeds on your branch but fails on mine?

veeceey:
@nodivbyzero Yes — consider this test case:

```go
type Inner struct {
    Name string
}
// A slice with nil and a zero-value struct
s := []*Inner{nil, {Name: ""}}
```

On my branch, this correctly passes the `unique` validation (nil and `&Inner{Name: ""})` are treated as distinct values). On your branch (#1532), both would map to `reflect.Zero(elem)`, so they'd collide and the validation would incorrectly fail, reporting a duplicate when there isn't one.

Happy to add this as an explicit test case if that would help illustrate the difference\!

nodivbyzero:
@veeceey something like this?
https://github.com/go-playground/validator/blob/26d790bd54e3b087aacf6144206d59be6e353839/validator_test.go#L11033-L11040

veeceey:
@nodivbyzero Yes, exactly like that! That test case would pass on my branch (nil and `&Inner{Name: ""}` are treated as distinct) but would fail on yours since both would map to the same zero-value key. Happy to add it as an explicit test case here if the maintainers would find that helpful for comparing the two approaches.

nodivbyzero:
> Yes, exactly like that! That test case would pass on my branch (nil and `&Inner{Name: ""}` are treated as distinct) but would fail on yours

@veeceey this test case has been added in my branch :)

veeceey:
Nice, glad that test case made it in! Sounds like #1532 has converged to handle the nil vs zero-value distinction properly now. Happy to close this one if maintainers prefer that approach — just let me know.

veeceey:
Closing in favor of #1532 which now covers the nil vs zero-value distinction as well. Thanks @nodivbyzero for picking it up!

Issue #1130, opened 2023-07-21
Cron validator accepting non conform cron string


github-charles-fraisse:
- [x] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#section-documentation) first?
- [x] I have looked at the examples provided that may showcase my question [here](/_examples)?

### Package version eg. v9, v10: 
v10.14.1

### Issue, Question or Enhancement:
Invalid string get accepted by the cron validator

### Code sample, to showcase or reproduce:
This field describe the cron input requested by my API (using fiber).
```go
	Cron string `json:"cron" validate:"required,cron"`
```
When sending the non valid "/15 * * * *" cron string to my API, the validator doesn't throw an error.

PR #1531, opened 2026-02-10
fixed validation of cron expression

### LINKED ISSUE #1529: [Bug]: Cron validator
### What happened?

Hello.

I was testing cron expressions using this library.

I encountered an error with the following schedules, which are semantically equivalent:

The schedule `10 0-3,9-23 * * *` is correct.

The schedule `10 9-23,0-3 * * *` is incorrect.

I don't understand why this library is generating this error.

I'm also curious if it's intentional.

### Version

v10

### Example Code

```go
type CronStruct struct {
  Cron string `validate:"cron"`
}

err = Valid.Struct(CronStruct)
```

thejus-r:
## Fixes Or Enhances

- fixes #1529 and #1130
- no validates range values & steps
- added new test cases

```go
// regexes.go
cronRegexString = `^(@(annually|yearly|monthly|weekly|daily|hourly|reboot)|@every (\d+(ns|us|µs|ms|s|m|h))+|(((\d+(-\d+)?|\*|\?|\d*L(?:W)?|\d*W|\d+#\d+|[A-Z]{3}(-[A-Z]{3})?)(\/\d+)?)(,(\d+(-\d+)?|\*|\?|\d*L(?:W)?|\d*W|\d+#\d+|[A-Z]{3}(-[A-Z]{3})?)(\/\d+)?)* ?){5,7})$`

```


zemzale (approved):
IMO this is better than what we have right now since it fixes the issues.

But this is still not perfect since as it would allow:
```
* 9-25 * * * 
```

Maybe update the test with something like this to have it documented?

```go

	t.Skip("FIXME: validation on minute/hour/day/month/weekday range is not working")

	failingTests := []struct {
		value    string `validate:"cron"`
		tag      string
		expected bool
	}{
		{"60 * * * *", "cron", false},
		{"* 25 * * *", "cron", false},
		{"* * 32 * *", "cron", false},
		{"* * * 13 *", "cron", false},
		{"* * * * 8", "cron", false},
	}

	for i, test := range failingTests {
		errs := validate.Var(test.value, test.tag)
		if test.expected {
			if !IsEqual(errs, nil) {
				t.Fatalf(`Index: %d cron "%s" failed Error: %s`, i, test.value, errs)
			}
		} else {
			if IsEqual(errs, nil) {
				t.Fatalf(`Index: %d cron "%s" should have errs`, i, test.value)
			}
		}
	}
```

PR #1536, opened 2026-02-28
fix(validation): Make `unix_addr` check useful (#1348)

### LINKED ISSUE #1348: bug: unix_addr check is useless

agentic-sanyam:
Hey folks,

This PR finally addresses #1348, where the `unix_addr` validation was, well, pretty much useless.

Turns out, `net.ResolveUnixAddr` for the "unix" network type in Go's standard library always succeeds (it just returns a `UnixAddr` struct with the provided name and network). This meant our `isUnixAddrResolvable` function always returned `true`, even for non-existent or invalid paths. Definitely not the behavior anyone would expect!

I've updated `isUnixAddrResolvable` to now leverage the existing `isUnixDomainSocketExists` function. This provides a much more robust check, ensuring the path actually exists and points to a valid Unix domain socket (or a Linux abstract socket).

Naturally, the `TestUnixAddrValidation` suite has been completely overhauled to reflect this new, stricter (and correct!) behavior. This includes:
*   Adding a skip for Windows, as Unix domain socket existence checks aren't really a thing there.
*   Dynamically creating a *real* Unix domain socket during the test to properly verify valid paths.
*   Expanding test cases for empty strings and non-existent paths, which now correctly fail.

Now `unix_addr` actually does what users expect it to do! 🎉

Cheers!

Fixes #1348


zemzale:
@go-playground/validator-maintainers this very much smells like a AI agent that just goes around and "fixes" random things. (new account creates like 4PRs in super unrelated repos, and username just implies it)
How we plan to deal with these kind of stuff? 

But in general can't review the PR since the included coverage report hangs github for me atm.

Issue #1496, opened 2025-11-21
Make usage of reflect.MethodByName() optional


vincentbernat:
Hey!

When using reflect.Method() or reflect.MethodByName() with a non-constant argument, dead code elimination is disabled for public functions. This makes the final binary bigger than it should. See for example https://github.com/kubernetes/kubernetes/issues/132216 for an effort on the subject.

`validatefn` validator uses such a method. Would it be OK to hide it behind a build constraint? Like `validator_no_reflectmethod`? This would prevent specifying a custom validator (a static one is still OK).

seanamos:
I also want to raise this issue. The impact of dead code elimination being disabled is not to be underestimated.
Where we are able to fix code/dependencies to get DCE enabled for large binaries, the savings can be _substantial_, 50%+-.

go-playground/validator is very popular, the go to and frequently recommended library for validation. Unfortunately anyone that imports it has DCE disabled, so the impact to the ecosystem is substantial. This is a hefty price to pay to support `validatefn` with a custom function name.

Unfortunately, I don't see a non-breaking way to do this in v10 without a build tag. 
`MethodByName` won't disable DCE if its argument is _constant_, but the current implementation allows you to specify a custom name.

References:
https://appliedgo.net/spotlight/reflection-binary-size/
https://github.com/golang/go/issues/62257

zemzale:
So yeah the only place where we use it in the `validateFn`. I can see us just adding an opt-in flag, to would cause the function to panic if used. 

On my windows machine I saw a simple validator go from 6.8MB to 5.4MB in size so we can easily remove the function with build flags. 

https://github.com/go-playground/validator/pull/1542 my POC here

PR #1542, opened 2026-03-14
Reduce build size with dead code elimination

zemzale:
Based on discussion in [1496](https://github.com/go-playground/validator/issues/1496) a experimental change to allow reducing the build sizes. 

This adds a build tag `validator_novalidatefn` that makes any usage of `validateFn` panic. The idea is that we can eliminate calls to `tryCallValidateFn` that in turn uses `field.MethodByName` that can't be dead code eliminated. 

Added a small block for docs, but I am not the best doc writer, so some feedback would be welcome there. Maybe something more is needed?

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -3,7 +3,6 @@ package validator
 import (
 	"bufio"
 	"bytes"
-	"cmp"
 	"context"
 	"crypto/sha256"
 	"encoding/hex"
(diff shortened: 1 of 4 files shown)


vincentbernat:
I think the build tag should be `validator_novalidatefn` to avoid collision.

sverdlov93:
We ran into this exact issue at JFrog while optimizing our Go service binary. Here's our analysis confirming the impact:

**Our setup:** ~1,480 packages, 99 MB binary (Linux amd64). We use `validator.New()` with custom tags but never use `validateFn`.

**What `whydeadcode` shows on our binary today:**
```
github.com/go-playground/validator/v10.tryCallValidateFn reachable from:
   github.com/go-playground/validator/v10.isValidateFn
   github.com/go-playground/validator/v10.isValidateFn·f
   github.com/go-playground/validator/v10.map.init.11
   github.com/go-playground/validator/v10.bakedInValidators
   github.com/go-playground/validator/v10.New
```

**Reviewed the code in this PR — it addresses the issue correctly.** The `MethodByName` call moves into `no_validate_fn.go` (default build), and `-tags validator_novalidatefn` compiles only the panic stub. The registration in `baked_in.go` stays in both paths, but the stub prevents `MethodByName` from appearing in the linker's reachability graph.

The build tag naming `validator_novalidatefn` is good — prefixed with the package name, consistent with how gin does `nomsgpack`. Would be great to see this land!

deankarn (approved):
LGTM

PR #1548, opened 2026-03-20
fix(fqdn): allow hyphens in last domain label

alihasan070707:
## Summary

The `fqdn` validator was rejecting valid domain names that contain hyphens in the last label (e.g. `test-site-http.test-site`), because the TLD character class `[a-zA-Z0-9]` did not include `-`.

## Change

One character added to the last label's character class in `fqdnRegexStringRFC1123`:

No other parts of the regex were modified.
 
## Test
 
Added regression case to [TestFQDNValidation](cci:1://file:///Users/admin/validator/validator_test.go:10628:0-10676:1):
 
- `test-site-http.test-site` → valid


alihasan070707:
@zemzale would it be possible for you to review the MR, we have a customer who is blocked due to this issue, would be great if you could review.

alihasan070707:
thank you @zemzale, do you know when would these changes will make it to the release?

PR #1560, opened 2026-04-15
fix: avoid max/min panic for struct-backed driver.Valuer fields

### LINKED ISSUE #1300: The go-playground/validator package panics when validating struct fields with the max tag on null.Int fields. This issue did not occur in earlier versions. Previously, the package would not panic.
- [ ] I have looked at the documentation [here](https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Maximum) first?
- [ ] I have looked at the examples provided that may showcase my question [here](/validator_test.go)?

v10.20.0

The go-playground/validator package panics when validating struct fields with the max tag on null.Int fields. This issue did not occur in earlier versions. Previously, the package would not panic.

```go
package main

import (
	"fmt"
	"gopkg.in/guregu/null.v4"
)

type User struct {
	Name     string      `json:"name"`
	Age      int         `json:"age"`
	NullAge  null.Int    `validate:"max=10"`
}

func main() {
	v := validator.New()
	user := User{
		Name:     "John Doe",
		Age:      30,
		NullAge:  null.IntFrom(15),
	}

	err := v.Struct(user)
	if err != nil {
		fmt.Println("Validation failed:", err)
	} else {
		fmt.Println("Validation succeeded")
	}
}
```

Expected behavior:

The validator should not panic when encountering the max tag on null.Int fields. Ideally, it should either validate the null.Int field properly or skip validation without causing a panic.

Actual behavior:

The validator panics when it encounters the max tag on null.Int fields.

Steps to reproduce:

	1.	Define a struct with a null.Int field and a max validation tag.
	2.	Create an instance of the struct with a value for the null.Int field.
	3.	Run the validator on the struct.

dapzthelegend:
## Fixes Or Enhances

Fixes #1300.

This change unwraps `database/sql/driver.Valuer` values during internal type extraction. That allows built-in validators like `max`/`min` to operate on the underlying scalar value instead of panicking on struct-backed valuer types (for example, `null.Int` and `sql.NullInt64`).

### What changed
- Added `driver.Valuer` unwrapping in `extractTypeInternal` (including pointer-receiver valuer support).
- Added regression test `TestMaxMinValidationWithDriverValuerStructFields` using `sql.NullInt64`.

### Verification
- `/tmp/go1.25.0/bin/go test ./... -run TestMaxMinValidationWithDriverValuerStructFields -count=1`
- `/tmp/go1.25.0/bin/go test ./... -run 'TestSQLValueValidation|TestSQLValue2Validation|TestValuerInterface' -count=1`

### DIFF
--- a/util.go
+++ b/util.go
@@ -1,6 +1,7 @@
 package validator
 
 import (
+	"database/sql/driver"
 	"fmt"
 	"reflect"
 	"regexp"
@@ -36,6 +37,10 @@ BEGIN:
 				goto BEGIN
 			}
 		}
+		if value, ok := extractDriverValuerValue(current); ok {
+			current = value
+			goto BEGIN
+		}
 
 		current = current.Elem()
 		goto BEGIN
@@ -53,6 +58,10 @@ BEGIN:
 				current = reflect.ValueOf(v.ValidatorValue())
 				goto BEGIN
 			}
+		}
+		if value, ok := extractDriverValuerValue(current); ok {
+			current = value
+			goto BEGIN
 		}
 
 		current = current.Elem()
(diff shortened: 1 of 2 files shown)


zemzale:
Thanks for the contribution, but this can already be done by:
- registering custom type function
- using `Valuer` interface

See: https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Valuer_Interface

PR #1562, opened 2026-04-15
fix: reject invalid dotted-decimal in hostname_rfc1123 validation

### LINKED ISSUE #1561: [Bug]: hostname_rfc1123 validation does not enforce IPv4 octet check
### What happened?

According to rfc1123: "a valid host name can never have the dotted-decimal form #.#.#.#, since at least the highest-level component label will be alphabetic."
However, for example, 277.168.0.1 is returned as a valid host, which is not a valid IPv4 address because of the first octet and should not be a valid hostname based on the rfc quoted above.

### Version

v10

### Example Code

```go
func TestHostnameRFC1123Validation(t *testing.T) {
	tests := []struct {
		param    string
		expected bool
	}{
		{"277.168.0.1", false},
         }
....
}

--- FAIL: TestHostnameRFC1123Validation (0.00s)
validator-master/validator_test.go:10679: Hostname: {277.168.0.1 false} failed Error: <nil>
FAIL
FAIL	github.com/go-playground/validator/v10	0.011s
FAIL
```

Herrtian:
## Summary

Fixes #1561

`hostname_rfc1123` validation accepted strings like `277.168.0.1` which are neither valid IPv4 addresses nor valid hostnames per RFC 1123.

RFC 1123 §2.1 states:
> "a valid host name can never have the dotted-decimal form #.#.#.#, since at least the highest-level component label will be alphabetic."

## Change

Added a check in `isHostnameRFC1123`: after the regex match, if the input looks like a dotted-decimal string (digits and dots only, with at least one dot), it must also be a valid IP address via `net.ParseIP`. Otherwise it's rejected.

| Input | Before | After |
|-------|--------|-------|
| `277.168.0.1` | pass | **fail** (invalid IP, looks like dotted-decimal) |
| `192.168.0.1` | pass | pass (valid IPv4) |
| `123.456.789.0` | pass | **fail** (invalid IP) |
| `example.com` | pass | pass (not dotted-decimal) |
| `1` | pass | pass (single label, not dotted-decimal) |

## Test plan

- [x] `go build ./...` passes
- [x] `TestHostnameRFC1123Validation` passes
- [x] `TestHostnameRFC1123AliasValidation` passes
- [x] Full test suite passes (all packages including translations)


zemzale (changes requested):
Thanks for the PR. 

As Copilot points out:
- the `looksLikeDottedDecimal` has a huge bug
- `net.ParseIP` gets called every time which is not needed for a LOT of cases

Also I would require some tests before accepting this change, that would cover all the cases that have been pointed out.

Herrtian:
Updated this to only apply the IP parse for four-part dotted decimal values, and added coverage for valid IPv4, invalid IPv4, and numeric hostname cases. go test ./... passes locally.

Issue #725, opened 2021-02-10
will not be invalid if hostname's last character is '-'


ophum:
### Package version eg. v9, v10: 
v10

### Issue, Question or Enhancement:
Question
use hostname rfc 1123, and when last character of hostname is '-', will not be invalid.
I think  invalid when last character `-`.

>    2.1  Host Names and Numbers
>
>      The syntax of a legal Internet host name was specified in RFC-952
>      [DNS:4]

in [https://tools.ietf.org/html/rfc1123#page-12](rfc1123 page-12)

> The last character must not be a minus sign or period. 

in [https://tools.ietf.org/html/rfc952]

### Code sample, to showcase or reproduce:

```go
package test

import (
	"testing"

	"github.com/go-playground/validator/v10"
)

type Test struct {
	Hostname string `validate:"hostname_rfc1123"`
}

func TestHostnameRFC1123(t *testing.T) {
	test := Test{
		Hostname: "test-",
	}
	expectErrString := "Key: 'Test.Hostname' Error:Field validation for 'Hostname' failed on the 'hostname_rfc1123' tag"
	validate := validator.New()
	if err := validate.Struct(&test); err != nil {
		if err.Error() != expectErrString {
			t.Errorf("expect:\t%s\nactual:\t%s\n", expectErrString, err.Error())
		} else {
			return
		}
	}
	t.Errorf("expect: not nil\nactual: nil\n")
}
```

```
$ go test -v
=== RUN   TestHostnameRFC1123
    hostname_test.go:26: expect: not nil
        actual: nil
--- FAIL: TestHostnameRFC1123 (0.00s)
```

ahmedkamalio:
@nodivbyzero this can be marked as fixed now!

ophum:
@ahmedkamalio thank you!!!!!!😍

PR #1565, opened 2026-04-20
feat: add origin validator for web origin URLs

ahmedkamalio:
## Fixes Or Enhances

Added `origin` tag for web origin URLs.

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -143,6 +143,7 @@ var (
 		"http_url":                      isHttpURL,
 		"https_url":                     isHttpsURL,
 		"uri":                           isURI,
+		"origin":                        isOrigin,
 		"urn_rfc2141":                   isUrnRFC2141, // RFC 2141
 		"file":                          isFile,
 		"filepath":                      isFilePath,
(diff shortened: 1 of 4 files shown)


baked_in.go
```diff
@@ -1561,6 +1562,66 @@ func isURI(fl FieldLevel) bool {
 	panic(fmt.Sprintf("Bad field type %s", field.Type()))
 }
 
+// isOrigin checks if a field value is a valid web origin URL with HTTP(S) scheme and host defined, but no path, query, or fragment.
+func isOrigin(fl FieldLevel) bool {
+	field := fl.Field()
+
+	switch field.Kind() {
```
nodivbyzero, line 1569: You only handle reflect.String, so a switch adds noise. A simple guard is clearer:
```
if field.Kind() != reflect.String {
    panic(...)
}
```

baked_in.go
```diff
@@ -1561,6 +1562,66 @@ func isURI(fl FieldLevel) bool {
...
+
+		if u.User != nil {
+			return false
+		}
+
+		hostname := u.Hostname()
+		if hostname == "" {
+			return false
+		}
+		if !hostnameRegexRFC1123().MatchString(hostname) && net.ParseIP(hostname) == nil {
```
nodivbyzero, line 1605: Flip condition order for cheaper checks. This avoids unnecessary regex work:

```suggestion
		if net.ParseIP(hostname) == nil && !hostnameRegexRFC1123().MatchString(hostname) {
```

validator_test.go
```diff
@@ -8791,6 +8791,133 @@ func TestUri(t *testing.T) {
 	PanicMatches(t, func() { _ = validate.Var(i, "uri") }, "Bad field type int")
 }
 
+func TestOrigin(t *testing.T) {
```
nodivbyzero, line 8794: there are a few test cases which won't pass:
```
		{"http://example-.com", false},
		{"http://-sub.example.com", false},
		{"http://sub-.example.com", false},
        {"http://example..com", false},
        {"http://example.com.", false},

```

ahmedkamalio:
- Fixed #725

nodivbyzero (approved):

PR #1570, opened 2026-04-28
feat: Add ISIN validator

walle250ai:
Adds an `isin` struct tag validator for International Securities Identification Numbers (ISO 6166). Validation checks:
1. Format: 2-letter country code + 9 alphanumeric chars + 1 check digit
2. Luhn checksum: letters are expanded to their two-digit values (A=10 … Z=35) before applying the standard Luhn algorithm.

- regexes.go: add isinRegexString and isinRegex
- baked_in.go: register "isin" tag and implement isISIN()
- validator_test.go: add TestISINStringValidation with valid ISINs (Apple, BP, BASF) and invalid cases (length, format, checksum, lowercase)

## Fixes Or Enhances

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -260,6 +260,7 @@ var (
 		"cron":                          isCron,
 		"spicedb":                       isSpiceDB,
 		"ein":                           isEIN,
+		"isin":                          isISIN,
 		"validateFn":                    isValidateFn,
 	}
 )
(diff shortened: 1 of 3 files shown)


zemzale (changes requested):
Thanks for the PR. 

The code looks good, but I would ask you to also update the docs before we can merge this.

PR #1577, opened 2026-05-15
fix(cron): anchor regex and accept full cron syntax

### LINKED ISSUE #1576: [Bug]: cron validator accepts arbitrary strings containing a cron-like substring
### What happened?

`cronRegexString` in `regexes.go` is the only full-string validator regex in the file without `^…$` anchors. `isCron` (`baked_in.go`) calls `cronRegex().MatchString(s)`, which matches a substring rather than the whole input — so any string that merely contains a cron expression passes validation.

### Version

v10.12.0+ including master

### Example Code

```go
Examples that currently (and incorrectly) pass `validate:"cron"`:

- "random text @daily more text"
- "x 1 2 3 4 5 y"
- "prefix @every 1h suffix"
- "not at all valid; trailing junk: * * * * *"
```

ahmedkamalio:
The cron validator's regex lacked `^...$` anchors, so any string containing a cron-like substring (e.g. `"random text @daily more text"`, `"prefix @every 1h suffix"`, `"x 1 2 3 4 5 y"`) was accepted as a valid cron expression.

Anchoring alone wasn't sufficient — the original field pattern only recognized plain numeric forms and was getting away with it via substring matching. Several existing positive test cases (`?`, `L`, `MON-FRI`, `#`, multi-field expressions with year) only passed because the regex matched a numeric fragment of the input. The field pattern is now expanded to accept the documented cron alphabet (`*`, `?`, digits, letters, `,`, `/`, `-`, `#`, `L`, `W`) while still rejecting bare single letters like "x" that aren't valid in any cron dialect.

Negative test cases added to lock in the new behavior, with a note documenting remaining false positives that would require a real cron parser to reject (out-of-range values, nonsense alphabetic tokens, etc).

- Fixes #1576

## Fixes Or Enhances


ahmedkamalio:
Failed lint is unrelated to the code touched in this PR

zemzale (approved):
Looks good, just rebase the PR please, the lint issue has been fixed

zemzale (approved):

PR #1578, opened 2026-05-16
fix: anchor cron regex so substrings can't pass validation

### LINKED ISSUE #1576: [Bug]: cron validator accepts arbitrary strings containing a cron-like substring

c-tonneslan:
## Fixes Or Enhances

Closes #1576

The `cron` validator regex is the only one in `regexes.go` without `^...$` anchors, so any string that contains a cron-like substring passes validation. Examples that currently pass `validate:\"cron\"` but shouldn't:

- `random text @daily more text`
- `x 1 2 3 4 5 y`
- `prefix @every 1h suffix`
- `not at all valid; trailing junk: * * * * *`
- `* * * * * trailing`
- `leading * * * * *`

Just adding anchors breaks several of the existing tests though, because they use quartz syntax (`?`, `L`, `MON-FRI`, `6#3`) that the old field pattern doesn't actually match. Those only passed before because `MatchString` could find a 5+ field substring of basic digits-and-stars somewhere inside the input. So I also widened the field character class to cover the characters the existing tests already expect (digits, `*`, `?`, `/`, `,`, `-`, `#`, and uppercase letters), which keeps the same set of expressions valid while rejecting leading/trailing junk.

The `@yearly`/`@daily`/`@every` branches are unchanged.


zemzale:
Closing this in favor of #1577

PR #1579, opened 2026-05-22
Expose errors returned by validateFn

### LINKED ISSUE #1575: [Bug]: `validateFn` does not propagage errors upwards
### What happened?

when using 

```go
package main

import (
	"errors"
	"fmt"

	val "github.com/go-playground/validator/v10"
)

type S struct {
	A string `validate:"validateFn"`
}

func (s *S) Validate() error {
	return errors.New("huha")
}

func main() {
	var s S
	e := val.New().Struct(&s)

	fmt.Printf("%v", e)
}

```

on a struct field `A`, the `Validate() error` function is called but the error returned from that function is never accesible in the result `e`.

Only an error:

```
Key: 'A' Error:Field validation for 'A' failed on the 'validateFn' tag"
```
is printed.

Also casting to `validation.ValidationError` does not help.

### Version

10.30

### Example Code

```go
see above.
```

puneetdixit200:
## Fixes Or Enhances

Fixes #1575.

`validateFn` methods that return a non-nil `error` now keep that error attached to the generated field validation error. Callers can inspect it with `errors.Is` or `errors.As` through either the returned `ValidationErrors` value or the individual `FieldError`.

This keeps the existing `FieldError` interface and validation failure shape unchanged while making the underlying `Validate() error` result accessible.

## Verification

- `GOMODCACHE=/tmp/validator-gomodcache GOCACHE=/tmp/validator-gocache go test ./... -run TestValidateFn -count=1`
- `GOMODCACHE=/tmp/validator-gomodcache GOCACHE=/tmp/validator-gocache go test -cover -race ./...`
- `PATH=/tmp/validator-bin:$PATH GOMODCACHE=/tmp/validator-gomodcache GOCACHE=/tmp/validator-gocache golangci-lint run`
- `git diff --check`

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -2,6 +2,7 @@ package validator
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -48,6 +49,20 @@ func (ve ValidationErrors) Error() string {
 	}
 
 	return strings.TrimSpace(buff.String())
+}
+
+// Unwrap returns the nested errors reported by individual field validations.
+func (ve ValidationErrors) Unwrap() []error {
+	errs := make([]error, 0, len(ve))
+	for _, fe := range ve {
+		if err := errors.Unwrap(fe); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	if len(errs) == 0 {
+		return nil
+	}
+	return errs
 }
 
 // Translate translates all of the ValidationErrors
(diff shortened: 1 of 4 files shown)


zemzale (approved):
Seems good.

jlakkist:
Thanks for implementing this feature, I need this for a feature in our project :+1: 

Any chance this could be merged soon?

hown3d:
This change does not print the error the underlying validateFn returned when using `fieldError.Error()`.
The Error function should include the newly added `fieldError.err` field in the `Error` method.

PR #1616, opened 2026-08-24
feat: Add hostname label validation tag

### LINKED ISSUE #1543: Enhancement: Add "dns_label" or "hostname_label" validator for single-label DNS validation
Currently, the hostname_rfc1123 tag validates a full hostname which can include multiple labels separated by dots (e.g., sub.domain.com). However, many modern use cases, such as validating a subdomain part in a multi-tenant application (e.g., [xxx].example.com) or Kubernetes resource names, require validating a single DNS label that strictly forbids dots.

**Proposed Change:**
Add a new baked-in validation tag, such as dns_label or hostname_rfc1123_label, that enforces the following RFC 1123 rules for a single part:

* Characters: Only alphanumeric (a-z, A-Z, 0-9) and hyphens (-).
* Start/End: Must start and end with an alphanumeric character.
* Length: 1 to 63 characters.
* No Dots: Strictly forbids the . character within the string.

```go
regexp=^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$
```

**Use Case:**
Validating a single subdomain prefix before appending it to a base domain in code:

```go
type Tenant struct {
    // Current hostname_rfc1123 would allow "illegal.sub.domain"
    // We need something that only allows "legal-sub"
    Subdomain string `validate:"required,dns_label"` 
}
```

KytnaCode:
## Fixes Or Enhances

Add `hostname_label` validation tag. Resolves #1543.

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -2916,7 +2916,48 @@ func isHostnameRFC1123(fl FieldLevel) bool {
 }
 
 func isHostnameLabel(fl FieldLevel) bool {
-	return hostnameLabelRegex().MatchString(fl.Field().String())
+	v := fl.Field().String()
+
+	const maxLabelLength = 63
+
+	// ASCII only.
+	isLetter := func(c byte) bool {
+		return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')
+	}
+
+	// ASCII only.
+	isDigit := func(c byte) bool {
+		return (c >= '0' && c <= '9')
+	}
+
+	valueLen := len(v)
+
+	// The empty label is not a valid label
+	if valueLen <= 0 || valueLen > maxLabelLength {
+		return false
+	}
+
+	firstChar := v[0]
+
+	if !isLetter(firstChar) && !isDigit(firstChar) {
+		return false
+	}
+
+	for i := range valueLen - 2 {
+		c := v[i+1]
+
+		if !isLetter(c) && !isDigit(c) && c != '-' {
+			return false
+		}
+	}
+
+	lastChar := v[valueLen-1]
+
+	if !isLetter(lastChar) && !isDigit(lastChar) {
+		return false
+	}
+
+	return true
 }
 
 func isFQDN(fl FieldLevel) bool {
--- a/baked_in.go
+++ b/baked_in.go
@@ -217,6 +217,7 @@ var (
 		"mac":                           isMAC,
 		"hostname":                      isHostnameRFC952,  // RFC 952
 		"hostname_rfc1123":              isHostnameRFC1123, // RFC 1123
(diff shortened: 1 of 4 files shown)


baked_in.go
```diff
@@ -2914,6 +2915,10 @@ func isHostnameRFC1123(fl FieldLevel) bool {
 	return hostnameRegexRFC1123().MatchString(fl.Field().String())
 }
 
+func isHostnameLabel(fl FieldLevel) bool {
+	return hostnameLabelRegex().MatchString(fl.Field().String())
```
nodivbyzero: Could we avoid the regex here? 
A small character-by-character check is simpler, allocation-free, and makes the RFC-style rules explicit.
The RFC 1123 rules are simple enough to express directly, which IMO makes the validation easier to read and maintain. 
The current regex looks correct, so this is just a suggestion rather than a blocker.

KytnaCode: Sure, I used a regex because most of the other rules use it, and I didn't want to add a more verbose check but I also prefer a character per character check

(Sorry for any typos, English is not my first language).

KytnaCode:
The hostname label rule uses local `isDigit` and `isLetter` functions that use ascii ranges for efficiency, I can change `alpha`, `alphanum`, `alphaspace` and `alphanumspace` to use them instead of regex, although, that doesn't work for unicode variants. I can open a pull request for that, what do you think?.

baked_in.go
```diff
@@ -2914,6 +2915,51 @@ func isHostnameRFC1123(fl FieldLevel) bool {
 	return hostnameRegexRFC1123().MatchString(fl.Field().String())
 }
 
+func isHostnameLabel(fl FieldLevel) bool {
```
nodivbyzero, line 2918: How about single-pass validation? 
```
func isHostnameLabel(fl FieldLevel) bool {
	v := fl.Field().String()
	n := len(v)

	// A valid hostname label must be between 1 and 63 characters long.
	if n == 0 || n > 63 {
		return false
	}

	for i := 0; i < n; i++ {
		c := v[i]
		
		// 1. Allow alphanumeric characters anywhere
		if (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') {
			continue
		}
		
		// 2. Allow hyphens only in the middle (not at the start or end)
		if c == '-' && i > 0 && i < n-1 {
			continue
		}
		
		// 3. Any other character, or a hyphen at the boundaries, is invalid
		return false
	}

	return true
}
```

KytnaCode: Ok, that's much better, the only thing I would change is to use a range over n instead of a traditional `for`.

PR #1621, opened 2026-09-03
Align excluded_without with its stated multi-field contract

dylanpulver:
## Fixes Or Enhances

`excludedWithout` passes the whole parameter to `requireCheckFieldKind` as one field name, so `excluded_without=Field2 Field3` looks up a field literally named `"Field2 Field3"`, never finds it, and falls back to `defaultNotFoundValue=true`. The field is then excluded **unconditionally**, whatever Field2 and Field3 hold — a silent wrong answer, not an error.

Its own doc comment says "when **any** of the other specified fields are not present", and all seven `excluded_with*`/`required_with*`/`required_without*` siblings parse with `parseOneOfParam2` and loop. `requiredWithout`, directly below it, got exactly this fix in #1422; the adjacent twin was not swept.

Measured with Field1 set:

| Field2 | Field3 | before | after |
|---|---|---|---|
| set | set | invalid | **valid** |
| set | absent | invalid | invalid |
| absent | set | invalid | invalid |
| absent | absent | invalid | invalid |

Single-parameter behaviour is unchanged (`parseOneOfParam2("A")` → `["A"]`).

Run back to back: upstream fails the new test, the fix passes, `go test ./...` is 25/25 ok. A naive variant using `excluded_without_all`'s all-absent loop also fails the new test — that case is what separates the two tags.

Drafted with Claude Opus 5 (`claude-opus-5`); every claim was executed.

### DIFF
--- a/baked_in.go
+++ b/baked_in.go
@@ -2253,8 +2253,11 @@ func requiredWithAll(fl FieldLevel) bool {
 // excludedWithout is the validation function
 // The field under validation must not be present or is empty when any of the other specified fields are not present.
 func excludedWithout(fl FieldLevel) bool {
-	if requireCheckFieldKind(fl, strings.TrimSpace(fl.Param()), true) {
-		return !hasValue(fl)
+	params := parseOneOfParam2(fl.Param())
+	for _, param := range params {
+		if requireCheckFieldKind(fl, param, true) {
+			return !hasValue(fl)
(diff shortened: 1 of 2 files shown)


nodivbyzero:
Thank you for your contribution! 
Could you please rebase your PR against the latest changes?

PR #1626, opened 2026-09-08
Accept hyphen grouping in credit_card validation

### LINKED ISSUE #1624: Should `credit_card` validator support dash-separated card numbers?
## What happened?

The `credit_card` validator currently only handles space-separated credit card numbers. Card numbers formatted with dashes (e.g., `4111-1111-1111-1111`) are incorrectly rejected because `digitsHaveLuhnChecksum` calls `strconv.Atoi` on `-` characters, which fails and returns `false`.

## Example Code

```go
package main

import ( 
    "fmt" 
)

type Payment struct { 
    CardNumber string `validate:"credit_card"` 
}

func main() { 
    validate := validator.New()

    // These are VALID card numbers but currently FAIL validation:
    payments := []string{
	    "4111-1111-1111-1111", // Visa, dash-separated
	    "3782-822463-10005",   // Amex, dash-separated
	    "4624-7482-3324-9780", // Discover, dash-separated
    }
    // These work correctly today:
    p := Payment{CardNumber: "4111 1111 1111 1111"} // space-separated
    fmt.Printf("Card: %s -> Error: %v\n", p.CardNumber, validate.Struct(p))
}
```

## Expected Behavior

Dash-separated card numbers should be validated the same as space-separated ones, since dashes are a legitimate card number format (used in Android payment forms, some banking apps, and recognized by ISO/IEC 7812).

## Actual Behavior

Dash-separated numbers are rejected as invalid.

## Version

v10.26.0

## Question

Would the maintainers accept a PR that strips dashes from the input before processing? The change would be a single `strings.ReplaceAll` call in `isCreditCard`, plus test cases. It wouldn't break any existing behavior since space-separated and plain-digit formats are already handled correctly.

official-burak:
## Fixes Or Enhances

credit_card already splits on spaces before the Luhn check, so 4624 7482 3324 9780 is valid, but the same digits with hyphens are not. Hyphens are now treated as the same grouping separator, so Visa 4-4-4-4 and Amex 4-6-5 work. Double hyphens still fail, matching the existing double-space rule.

Fixes #1624


baked_in.go
```diff
@@ -3468,7 +3468,9 @@ func isSpiceDB(fl FieldLevel) bool {
 
 // isCreditCard is the validation function for validating if the current field's value is a valid credit card number
 func isCreditCard(fl FieldLevel) bool {
-	val := fl.Field().String()
+	// Hyphens are a common grouping separator (Visa 4-4-4-4, Amex 4-6-5).
+	// Treat them like spaces so the existing segment rules still apply.
+	val := strings.ReplaceAll(fl.Field().String(), "-", " ")
```
nodivbyzero, line 3473: Using strings.ReplaceAll, strings.Split, bytes.Buffer, and strings.Split(..., "") creates multiple unnecessary string and slice allocations. 

It would be nice to see a version which processes the string in a single pass, allocates memory only once, validates that only digits/spaces/hyphens are present and then uses Luhn checker.

PR #1633, opened 2026-09-15
fix: allow signed integer fields in port validation

### LINKED ISSUE #1632: Allow `port` to be int
As now, the `port` validator supports only `uint` fields.

https://github.com/go-playground/validator/blob/dfe35cf8317892133dfe31e36054dcbf36aab604/baked_in.go#L3047-L3051

However, the stdlib return ports as `int` most of the time. E.g.:
- https://pkg.go.dev/net#LookupPort
- https://pkg.go.dev/net#TCPAddr
- https://pkg.go.dev/net#UDPAddr

Also, converting an `uint` to string requires either to convert it to `uint64` (`strconv.FormatUint`) or to `int` (`strconv.Itoa`).

Both `int` and `uint` should be supported for `port`.

Related to:
- https://github.com/go-playground/validator/issues/1216
- https://github.com/go-playground/validator/pull/1294

EJJI0113:
## Fixes Or Enhances

The `port` validator panics on signed integer fields, including a valid `int(8080)`. This prevents validating port values using the types used by Go's standard networking APIs.

Handle signed integer kinds with `reflect.Value.Int()` while preserving unsigned integer validation and the existing 1–65535 range. Extend the regression tests to cover signed kinds, named integer types, negative and out-of-range values, and unsigned compatibility. Update the port documentation accordingly.

Validation:
- `go test . -run '^Test_port_validator$' -count=1`
- `go test -race -covermode=atomic -coverprofile=../validator-coverage.out ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.13.1 run`

Fixes #1632.


baked_in.go
```diff
@@ -3045,9 +3045,16 @@ func isHostnamePort(fl FieldLevel) bool {
...
+	field := fl.Field()
 
-	return val >= 1 && val <= 65535
+	switch field.Kind() {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		val := field.Int()
+		return val >= 1 && val <= 65535
+	default:
+		val := field.Uint()
+		return val >= 1 && val <= 65535
```
ilovelinux: I'd implement the same logic implemented in `isIso3166AlphaNumeric` and similar functions. The panic error is going to be more explicit.

```suggestion
	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
		val := field.Uint()
		return val >= 1 && val <= 65535
	default:
		panic(fmt.Sprintf("Bad field type %s", field.Type()))
```

ilovelinux (approved):
LGTM, a minor suggestion to explicitly reject unsupported values.

ilovelinux (approved):
LGTM

doc.go
```diff
@@ -1445,7 +1445,7 @@ can be used to validate fields typically passed to sockets and connections.
 
 # Port
 
-This validates that the value falls within the valid port number range of 1 to 65,535.
+This validates that a signed or unsigned integer value falls within the valid port number range of 1 to 65,535.
```
nodivbyzero:
```suggestion
This validates that an integer value (signed or unsigned) falls within the valid port range of 1-65,535.
```

nodivbyzero (dismissed):
Thanks for contributing! The changes look good to me with just one minor doc change request.

nodivbyzero (approved):

PR #1642, opened 2026-09-21
fix: accept Taiwan six-digit postcodes

### LINKED ISSUE #1641: [Bug]: post_code validation rejects Taiwan 6-digit (3+3) postal codes
### Summary
Requests with Taiwan (TW) addresses fail validation when post_code uses the newer 6-digit format (3+3), even though this format is valid per Chunghwa Post.

https://en.wikipedia.org/wiki/Postal_codes_in_Taiwan#:~:text=On%20March%203%2C%202020%2C%20the%20postal%20codes%20are%20further%20extends%20to%20three%20plus%20three%20(3%2B3)%20numeral%20digits

### Expected behavior
6-digit Taiwan postcodes (3+3) should be accepted, since they are the current official format introduced by Chunghwa Post in 2020.

### Version

v10.15.1 but this is also reproducible on latest v10.30.5

### Example Code

```go
package main

import (
	"fmt"
)

func main() {
	v := validator.New()
	codes := []string{"100", "10001", "100001", "106409", "10617", "106170"}
	fmt.Println("validator v10.30.5")
	for _, code := range codes {
		err := v.Var(code, "postcode_iso3166_alpha2=TW")
		fmt.Printf("  %q => %s\n", code, map[bool]string{true: "valid", false: "invalid"}[err == nil])
	}
}
```

sb123sb123:
Fixes #1641

## Problem

The `postcode_iso3166_alpha2=TW` validator rejects Taiwan's current six-digit `3+3` postal-code format.

## Cause

The Taiwan pattern accepted only three digits or three digits followed by exactly two more digits.

## Fix

Allow the existing three- and five-digit forms plus the six-digit `3+3` form, with regression coverage for a valid six-digit code and an overlong seven-digit value.

## Tests

- `go test . -run TestPostCodeByIso3166Alpha2 -count=1`
- `go test . -count=1`
- `go test ./... -count=1`
- `go vet ./...`
- normalized `gofmt -d` checks for changed Go files
- `git diff --check`

## Limitations

Tests ran on the requested Windows Git Bash host with Go 1.26.7. The host could not reach `proxy.golang.org`, so Go modules were downloaded through the reachable `goproxy.cn`; no platform-specific CI result is claimed here.

## AI assistance disclosure

This PR was prepared with AI assistance from GPT-5 (OpenAI Codex). AI assistance covered issue triage, the minimal code and regression-test changes, and validation; the submitted diff and test results were reviewed before submission.


ajwad-shaikh (approved):

validator_test.go
```diff
@@ -14855,6 +14855,10 @@ func TestPostCodeByIso3166Alpha2(t *testing.T) {
 			{"00803", true},
 			{"1234567", false},
 		},
+		"TW": {
+			{"106409", true},
+			{"1234567", false},
```
nodivbyzero:
```suggestion
		{"100", true},       // 3 digits (Valid)
		{"10042", true},     // 5 digits (Valid)
		{"106409", true},    // 6 digits (Valid)
		{"040", true},       // Leading zero (Valid)
		{"12", false},       // Too short
		{"1234", false},     // 4 digits (Invalid)
		{"1234567", false},
		{"100-42", false},   // Contains hyphen (Invalid for strict numeric check)
		{"abcdef", false},   // Letters (Invalid)
```

